PR #737, opened 2022-02-25
fix(editing): Articles not having topics assigned to them [BACK-1335]

nina-py:
## Goal

Fix a bug where for some prospects (initially thought to be syndicated articles) the topic appeared to be saved when the prospect was recommended but it wasn't showing up on the card of this item on the Corpus page.

Narrowed down the initial bug report to multi-word topic names: instead of saving `HEALTH_FITNESS` on the backend, for example, the frontend was sending through `HEALTH & FITNESS` to the API, capitalising the display name instead of using the enum code from the shared data.

- Updated the `topic` select in the `ApprovedItemForm` component, and made similar changes to the `language` and `curationStatus` selects. Updated tests to check both option values and display values for all three fields.

## Reference

- https://getpocket.atlassian.net/browse/BACK-1335

### DIFF
--- a/src/curated-corpus/helpers/helperFunctions.ts
+++ b/src/curated-corpus/helpers/helperFunctions.ts
@@ -8,21 +8,6 @@ import {
   Prospect,
   UrlMetadata,
 } from '../../api/generatedTypes';
-/**
- *
- * This is a helper file that contains some helper functions. Right now
- * it only has one but we could add more to it
- */
-/**
- *
- * This is a helper file that contains some helper functions. Right now
- * it only has one but we could add more to it
- */
-/**
- *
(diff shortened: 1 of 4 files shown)


src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -8,21 +8,6 @@ import {
...
-/**
- *
- * This is a helper file that contains some helper functions. Right now
- * it only has one but we could add more to it
- */
-/**
- *
- * This is a helper file that contains some helper functions. Right now
- * it only has one but we could add more to it
- */
```
nina-py, line 11: Unrelated to the main ticket - deleting this set of comments in the helper functions file which has half a dozen helper functions by now.

PR #753, opened 2022-03-03
fix(collections): add 'target=_blank' to story links

nina-py:
## Goal

As discussed in Slack - collection stories opening by default in the same tab lead curators away from the single-page app unintentionally.

- Added `target="_blank" to links on the collection story card. Updated tests. ESLint also made me add `rel="noreferrer"` as that apparently mitigates a security risk in older browsers.

Not ticketed in JIRA.


jpetto (approved):
love these attention to detail fixes 😄

PR #754, opened 2022-03-03
fix(prospecting): Use display names for topics instead of codes with underscores

nina-py:
## Goal

Achieve consistency throughout the curated corpus app with the way topics are displayed.

- Added a helper function that converts topic codes into display names (used
in the topic dropdown), it also provides the `N/A` option for prospects without
a topic set.

- Now using this helper on every content card - approved item card, mini schedule
card and prospect card to show topic names consistently.

- Added and updated unit tests where necessary

## Reference

Tickets:

- Not ticketed, this has been bugging me for a while (sorry @bassrock 😄)

### DIFF
--- a/src/curated-corpus/components/MiniScheduleCard/MiniScheduleCard.tsx
+++ b/src/curated-corpus/components/MiniScheduleCard/MiniScheduleCard.tsx
@@ -14,6 +14,7 @@ import LabelOutlinedIcon from '@material-ui/icons/LabelOutlined';
 import CheckCircleOutlineIcon from '@material-ui/icons/CheckCircleOutline';
 import BookmarksIcon from '@material-ui/icons/Bookmarks';
 import { useStyles } from './MiniScheduleCard.styles';
+import { getDisplayTopic } from '../../helpers/helperFunctions';
 
 interface MiniScheduleCardProps {
   item: ScheduledCuratedCorpusItem;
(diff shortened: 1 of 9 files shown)


src/curated-corpus/components/MiniScheduleCard/MiniScheduleCard.tsx
```diff
@@ -59,30 +60,23 @@ export const MiniScheduleCard: React.FC<MiniScheduleCardProps> = (
...
           {item.approvedItem.isSyndicated && (
             <ListItem disableGutters>
               <ListItemIcon className={classes.listItemIcon}>
                 <CheckCircleOutlineIcon fontSize="small" />
               </ListItemIcon>
-              <ListItemText
-                className={classes.listItemText}
-                secondary={'Syndicated'}
-              />
+              <ListItemText secondary={'Syndicated'} />
```
nina-py, line 71: Also removed a CSS style that was capitalising the first letter of the topic - this is no longer needed. Before this change, the component would take the topic code - all uppercase - for example, `TECHNOLOGY`, transform it into lowercase with JS and capitalise the first letter with CSS 🤦. It worked when topic codes in the shared data had symbols and spaces, e.g. `HEALTH & FITNESS`, but definitely didn't when the shared data was updated to use underscores instead, e.g. `HEALTH_FITNESS`.

jpetto (approved):
this is lovely polish! 💅

PR #757, opened 2022-03-03
feat(surface select): auto-select surface if only one is available

jpetto:
## Goal

when scheduling an item and having access to only one scheduled surface, auto-select that surface.

- set id to an empty string for manually added prospects
- do not call prospect api after manually adding a prospect
- add some tests

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1336
- https://getpocket.atlassian.net/browse/BACK-1357

### DIFF
--- a/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.test.tsx
+++ b/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.test.tsx
@@ -13,7 +13,7 @@ describe('The ScheduleItemForm component', () => {
   const scheduledSurfaces: ScheduledSurface[] = [
     {
       name: 'en-US',
-      guid: 'EN_US',
+      guid: 'NEW_TAB_EN_US',
       utcOffset: -4000,
       prospectTypes: [
         ProspectType.Global,
@@ -23,7 +23,7 @@ describe('The ScheduleItemForm component', () => {
     },
     {
       name: 'de-DE',
-      guid: 'DE_DE',
+      guid: 'NEW_TAB_DE_DE',
       utcOffset: 1000,
       prospectTypes: [ProspectType.Global],
     },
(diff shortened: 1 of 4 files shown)


src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -345,37 +345,66 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
...
   };
 
+  /**
+   * common stuff we (may) need to do after creating an approved item
+   *
+   * @param approvedItemStatus (CuratedStatus): the status of the approved item
+   * @param approvedItemData (any): the data we get back from the createApprovedItem mutation
+   * @param filterProspects (boolean): whether or not we need to filter the list of prospects on the screen
+   */
+  const postCreateApprovedItem = (
```
jpetto, line 387: do we like this abstraction? i don't love it, but it does the job. feels like a bit more work to fully untangle.

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -345,37 +345,66 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
...
+    approvedItemStatus: CuratedStatus,
+    approvedItemData: any,
```
jpetto, line 389: is there a better type i can give this? i couldn't figure out how to tell what this is exactly.

kkyeboah (approved):
Looks good to me, tests made it easy to reason about the changes, thanks for adding those.

src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.test.tsx
```diff
@@ -66,4 +66,84 @@ describe('The ScheduleItemForm component', () => {
...
+          data-testId="surface-selector"
+          handleDateChange={jest.fn()}
+          lookupCopy=""
+          selectedDate={DateTime.local()}
+          onSubmit={handleSubmit}
+          scheduledSurfaces={scheduledSurfaces}
+          scheduledSurfaceGuid="NEW_TAB_EN_US"
+          approvedItemExternalId={'123abc'}
+        />
+      </MuiPickersUtilsProvider>
```
nina-py, line 109: @jpetto , I'm curious why you add `data-testId` to the component in the new tests but then don't query the DOM by this prop?

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
nina-py: "Untangle" is the key word when it comes to page-level components in this codebase 🙃.

nina-py (approved):

src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.test.tsx
jpetto: oh! yeah, that `data-testId` isn't needed. at one point i was trying to find the react equivalent of `getElementById` and thought that might do it. (spoiler: in this case, it doesn't.) i just forgot to remove that attribute when i moved to `getByLabelText`.

jpetto: but now that i think about it, maybe `data-testId` would work - at that point i also didn't realize i had to cast the result `as HTMLSelectElement`.

PR #759, opened 2022-03-04
feat(prospecting): [BACK-1338] When prospecting, jump straight to scheduling if corpus item already exists

nina-py:
## Goal

Instead of showing an error if a prospect has already been curated, jump straight to scheduling if the "Recommend" button is used.

- [x] This initial commit implements this workflow for manually adding an item.
- [x] The main prospecting workflow also needs to be updated.
- [x] Additionally, when you try to add an item to the corpus that is in the rejected corpus, you get an INTERNAL_SERVER_ERROR. we should display a helpful error message instead - this is a backend issue and needs its own PR in `curated-corpus-api` (https://github.com/Pocket/curated-corpus-api/pull/600)

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1338


nina-py:
Closing in favour of #783

PR #782, opened 2022-03-23
chore(language): Update language field mapping for prospect item

Herraj:
## Goal

Update `language` field mapping for prospect item
**aside:** add scheduledSurfaceGuid field to getScheduledItems query

Tickets:

- N/A

### DIFF
--- a/src/curated-corpus/components/ApprovedItemForm/ApprovedItemForm.tsx
+++ b/src/curated-corpus/components/ApprovedItemForm/ApprovedItemForm.tsx
@@ -6,6 +6,7 @@ import {
   CuratedStatus,
 } from '../../../api/generatedTypes';
 import {
+  ApprovedItemFromProspect,
   curationStatusOptions,
   DropdownOption,
   languages,
@@ -32,7 +33,7 @@ interface ApprovedItemFormProps {
   /**
    * The approved item that needs to be edited.
    */
-  approvedItem: ApprovedCuratedCorpusItem;
+  approvedItem: ApprovedCuratedCorpusItem | ApprovedItemFromProspect;
 
   /**
    * On submit handle function called on the 'Save' button click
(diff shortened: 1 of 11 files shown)


src/curated-corpus/components/ApprovedItemForm/ApprovedItemForm.tsx
```diff
@@ -32,7 +33,7 @@ interface ApprovedItemFormProps {
   /**
    * The approved item that needs to be edited.
    */
-  approvedItem: ApprovedCuratedCorpusItem;
+  approvedItem: ApprovedCuratedCorpusItem | ApprovedItemFromProspect;
```
Herraj, line 36: I've had to create an new type as you'll see below and use it here and also in the `ImageUpload` component. 

This is where we start seeing some effects of coupling since a single component (`ApprovedItemForm`) is being used to accommodate different items/types now. I think it is fine since the other solution of copying this component for a prospect type would be too redundant since the UI would be the exact same.

Herraj (commented):
Let me know your thoughts on the approach/any red flags you see

src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -55,6 +53,14 @@ export const transformProspectToApprovedItem = (
 export const transformUrlMetaDataToProspect = (
   metadata: UrlMetadata
 ): Prospect => {
+  // set language to undefined if metadata.language is an empty string or undefined.
+  // if not, then map it from string to its corresponding CorpusLanguage enum value
+  const language = !metadata.language
+    ? undefined
+    : metadata.language === 'en'
+    ? CorpusLanguage.En
+    : CorpusLanguage.De;
```
Herraj, line 62: This might be too confusing, I can break it down using an `if()` 😬

src/curated-corpus/components/ApprovedItemForm/ApprovedItemForm.tsx
nina-py: I'm happy with this! All it does is swap the type for the `language` prop for something compatible with the data we get, it's the right approach.

src/curated-corpus/helpers/helperFunctions.ts
nina-py: Maybe a comment? In a follow-up PR perhaps as I'd really like to merge this one and make it easier for myself in the next frontend branch :)

nina-py (approved):

PR #786, opened 2022-03-29
fix(corpus): [BACK-1380] Cannot edit items in the corpus - language enum error

nina-py:
## Goal

Allow curators to edit approved items again.

- Removed string transformations that are no longer necessary to conform
to mutation input requirements and that were the cause of the bug.

- Updated shared data to use enums.

## Reference

Tickets:

-https://getpocket.atlassian.net/browse/BACK-1380


jpetto (approved):
rollin'!

PR #789, opened 2022-03-29
refactor: [BACK-1377] Remove `Curated` from query/mutation names

nina-py:
## Goal

Update the frontend in tandem with https://github.com/Pocket/curated-corpus-api/pull/650 and https://github.com/Pocket/prospect-api/pull/399. 

- Regenerated types based on the GraphQL schema from Admin API

- Updated mutations and queries to use the new names.

- Updated all components and tests that were referencing the generated types.

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1377


jpetto (approved):
🔗

PR #802, opened 2022-04-01
chore(refactor): remove unecessary guid to timezone map and fiz z-index on new tab selector

Herraj:
## Goal

Remove the `guidToTimezone` map structure in favour of the newly added `ianaTimezone` field on the scheduled item received from the query

### DIFF
--- a/src/curated-corpus/helpers/definitions.ts
+++ b/src/curated-corpus/helpers/definitions.ts
@@ -67,31 +67,3 @@ export const curationStatusOptions: DropdownOption[] = [
 export type ApprovedItemFromProspect = Omit<ApprovedCorpusItem, 'language'> & {
   language: CorpusLanguage | undefined;
 };
-
-// scheduled surface guid to timezone map
-export const guidToUtcOffset = [
-  {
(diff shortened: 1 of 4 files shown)


src/curated-corpus/helpers/definitions.ts
```diff
@@ -67,31 +67,3 @@ export const curationStatusOptions: DropdownOption[] = [
...
-
-// scheduled surface guid to timezone map
-export const guidToUtcOffset = [
-  {
-    guid: 'NEW_TAB_EN_US',
-    timeZone: 'America/New_York',
-  },
-  {
-    guid: 'NEW_TAB_DE_DE',
-    timeZone: 'Europe/Berlin',
```
sri-kirsh, line 79: are we removing these coz its handled in the server now, and client doesn't need to maintain these mapping? 

i remember an old PR in the server where we had explicit mapping from surfaceGuid to timezone: https://github.com/Pocket/curated-corpus-api/pull/661/files

sri-kirsh (approved):

jpetto (approved):
yay cleanup!

PR #811, opened 2022-04-14
feat(scheduling): [BACK-1403] Implement "Move to bottom" button on Schedule page

nina-py:
## Goal

- Added a "Move to bottom" button to each scheduled item - this button
allows curators to move the item to the bottom of the list of scheduled
items consumed by client applications. This will enable predictable
story order on Pocket Hits surfaces.

- Updated tests.

## I'd love feedback on...

- The copy on the button. Is there something more succinct? The long copy gets the layout out of whack. Could we get away with just "Reorder"?

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1403

## Video walkthrough

https://user-images.githubusercontent.com/22447785/163322032-aa08dafc-ee04-4919-8eac-257a9f4a4523.mov


jpetto (approved):
i think this is great and a very elegant solution to the problem.

i also don't love how "move to bottom" breaks our layout, but i don't see why that has to block anything. we can perhaps find better language later on.

the curators in slack seem to like it, too. nice work nina! 🏆

PR #824, opened 2022-05-02
chore(generatedTypes): update generated types

jpetto:
## Goal

just updates generated types based on recent API update


bassrock (approved):
Yikes a big pr here.

PR #837, opened 2022-05-31
fix(manual schedule): [BACK-1456] Fix manual scheduling throwing an error

Herraj:
## Goal

After adding a manual prospect/item and then trying to schedule it, users were getting an error toast even though the item was being successfully scheduled. This change fixes that.

Tickets:

- [BACK-1456]

[BACK-1456]: https://getpocket.atlassian.net/browse/BACK-1456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
+++ b/src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
@@ -515,10 +515,10 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
         // Hide the loading indicator
         formikHelpers.setSubmitting(false);
 
-        // In case this prospect already existed in the corpus database,
+        // In case this prospect already existed in the corpus database i.e it has an approvedItem on it,
         // mark it as curated at this point
-        // A manually added item (currentProspect) won't have an id so we skip this check and mutation call
-        if (currentProspect?.id && approvedItem?.url === currentProspect?.url) {
+        // A manually added item/currentProspect won't have an approvedItem on it so this check is never hit
+        if (currentProspect?.approvedCorpusItem) {
           runMutation(updateProspectAsCurated, {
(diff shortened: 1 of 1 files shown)


src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -515,9 +515,10 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
         // Hide the loading indicator
         formikHelpers.setSubmitting(false);
 
-        // In case this prospect already existed in the corpus database,
+        // In case this prospect already existed in the corpus database i.e it has an approvedItem on it,
         // mark it as curated at this point
-        if (approvedItem?.url === currentProspect?.url) {
+        // A manually added item/currentProspect won't have an approvedItem on it so this check is never hit
+        if (currentProspect?.approvedCorpusItem) {
```
Herraj, line 521: Originally, we weren't resolving an `approvedItem` on the `Prospect` type in the graph. So we had two state variables that were being set and their urls were being compared to see if they were the same item/if it already existed in the corpus DB.

This lead to manually added items to also pass this check but not have an `id` since the item didn't come through the ML job/prospect-api, hence the error was thrown for marking it as curated but the initial schedule mutation passed. 

Now we can simply check if the prospect has an `approvedItem` on it to verify that is in the Corpus DB already and mark it as curated when an user schedules it. Note we already use this logic below to render the `<ExistingProspectCard>` component.

jpetto: thank you for this explanation!

jpetto (approved):

PR #840, opened 2022-06-23
fix(mozilla): moving to mozilla auth proxy

bassrock:
## Goal

Move to Mozilla Auth Proxy instead of our older cognito proxy. this should ameliorate the problem we've faced for years when folks are in too many mozilla groups and access breaks (due to the group size breaking the auth payload size restrictions).

- https://getpocket.atlassian.net/browse/BACK-1505

### Note

the client id value in your local `.env` file will need to be updated! ask in slack (or check AWS cognito) for the new value!

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -15,15 +15,15 @@ export const config = {
   oauth2: {
     provider: `${
       process.env.REACT_APP_OAUTH2_PROVIDER ??
-      'https://pocket-admin-prod.auth.us-east-1.amazoncognito.com/oauth2'
+      'https://mozilla-auth-proxy.getpocket.com/oauth2'
     }`,
     // So is the client ID.
     clientId: `${
-      process.env.REACT_APP_OAUTH2_CLIENT_ID ?? '6qt94s9d651k24mvul9q2lbrcv'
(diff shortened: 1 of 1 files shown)


jpetto:
did a push to dev and am getting the following error (in a private window with a fresh SSO login):

```
message: "Invalid token specified"
​
stack: "@https://curation-admin-tools.getpocket.dev/static/js/2.3a6bd141.chunk.js:2:956002\ni@https://curation-admin-tools.getpocket.dev/:1:1323\n@https://curation-admin-tools.getpocket.dev/static/js/2.3a6bd141.chunk.js:2:371495\n@https://curation-admin-tools.getpocket.dev/static/js/2.3a6bd141.chunk.js:2:377607\ni@https://curation-admin-tools.getpocket.dev/:1:1323\n833@https://curation-admin-tools.getpocket.dev/static/js/main.2669d016.chunk.js:1:12761\ni@https://curation-admin-tools.getpocket.dev/:1:1323\nr@https://curation-admin-tools.getpocket.dev/:1:1192\nt@https://curation-admin-tools.getpocket.dev/:1:1055\n@https://curation-admin-tools.getpocket.dev/static/js/main.2669d016.chunk.js:1:89\n"
```

will try to dig in a bit after this work week.

jpetto:
just pushed this to dev again and it is working!

katerinachinnappan (approved):

PR #847, opened 2022-06-24
organize dependencies

Herraj:
## Goal

Move some devDependencies out of the Dependencies. Tested on DEV.

Tickets:

- [BACK-1507]

[BACK-1507]: https://getpocket.atlassian.net/browse/BACK-1507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -40,11 +26,19 @@
...
     "react-oauth2-pkce": "2.0.7",
     "react-router-dom": "5.3.3",
+    "@types/react-router-dom": "5.3.3",
     "react-scripts": "4.0.3",
     "slugify": "1.6.5",
     "typescript": "4.5.5",
     "web-vitals": "2.1.4",
-    "yup": "0.32.11"
+    "yup": "0.32.11",
+    "@types/apollo-upload-client": "17.0.0",
```
bassrock, line 35: Can't the types packages go in the dev depependencies?

Herraj: I tried moving these into dev dependencies but seems like their corresponding non @types packages need to import from them. 

I'll give it another go and see if it compiles and builds a docker image!

Herraj: So I played around with it, one of the approach was to add a `noImplicityAny: false` flag to the `tsconfig` but we shouldn't really be doing that. 

I tried to build it locally and figured out the `@types/` packages that were needed for the app dependencies. The rest I've moved to devDependencies!

jpetto (approved):
this is nice cleanup!

my guess is the type packages required for prod are so typescript can build?

PR #854, opened 2022-07-07
chore(generatedTypes): guarantee authors data

jpetto:
## Goal

update generated types to guarantee authors data.

(do we need any actual code changes for this? maybe not?)

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1497

### DIFF
--- a/src/_shared/utils/flattenAuthors.ts
+++ b/src/_shared/utils/flattenAuthors.ts
@@ -2,7 +2,6 @@ import {
   CollectionAuthor,
   CollectionStoryAuthor,
   CorpusItemAuthor,
-  Maybe,
 } from '../../api/generatedTypes';
 
 /**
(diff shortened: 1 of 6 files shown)


nina-py (commented):
If authors are now supplied for every curated item, we could get rid of the code that steps around the no-authors issue here: https://github.com/Pocket/curation-admin-tools/blob/f1df46e8af65c62558256f5445015b9b01792a5c/src/_shared/utils/flattenAuthors.ts#L18.

Additionally, the `ApprovedItemListCard` component shouldn't need the "N/A" display option any more: https://github.com/Pocket/curation-admin-tools/blob/f1df46e8af65c62558256f5445015b9b01792a5c/src/curated-corpus/components/ApprovedItemListCard/ApprovedItemListCard.tsx#L77

jpetto:
@nina-py thank you! i think i've got it all cleaned up?

nina-py (approved):
🚀

PR #861, opened 2022-07-12
chore(integration tests): [BACK-1511] Add integration tests for `ScheduleItemFormConnector`

Herraj:
## Goal

Develop integration tests strategy and apply to existing component.

- Added integration tests for ScheduleItemFormConnector that test the scheduled
surface dropdown that takes its data from the `getScheduledSurfacesForUser` query.

- Added several mocks for scheduled surfaces and scheduled items that can be reused
by other integration tests.

Tickets:

- [Link to JIRA ticket](https://getpocket.atlassian.net/browse/BACK-1511)

## Implementation Decisions
This might change depending on what team decides on using whether it is apollo's `MockedProvider` or `msw` library (NP - taking the pragmatic approach of setting `msw` aside for now)

NP - note the scope of this ticket/PR has been trimmed down somewhat. The second query in the component is as yet untested - we should break up this component or figure out how to get a test working within the existing suite (see the last test in the test suite).


src/curated-corpus/components/ScheduleItemFormConnector/ScheduleItemFormConnector.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { DateTime } from 'luxon';
+
+describe('ScheduleItemFormConnector', () => {
+  let mocks = [];
+
+  it('loads the form with all scheduled surfaces', async () => {
+    mocks = [mock_AllScheduledSurfaces];
+
+    render(
+      <MockedProvider mocks={mocks}>
```
jpetto, line 24: was poking around in [the apollo docs](https://www.apollographql.com/docs/react/development-testing/testing/#setting-addtypename) this morning and saw they recommend adding `addTypename={false}` to these `<MockedProvider>` components. these tests seem to be working fine as-is though, so...?

jpetto (approved):
i think the pattern being established here is solid. i like the mock helper quite a bit. it does seem like the issue raised here is how to properly construct this connector component - but that's outside the scope of this ticket.

i say we go with this and make component updates in other tickets.

nice job both!

PR #881, opened 2022-07-18
chore(tests): [BACK-1535] Add unit tests for ScheduleItemModal component

nina-py:
## Goal

Improve test coverage of our frontend code. Note: no change to functionality, tests added/amended only.

- Added the missing unit tests.

- Added comments to component props.

- Fixed up test mocks elsewhere whenever the IDE highlighted missing props.

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1535

PR #882, opened 2022-07-18
chore(tests): [BACK-1513] Add tests to RejectedPage component

nina-py:
## Goal

Have test coverage for our most important components - pages! Starting with the easiest page of all, the rejected corpus items page.

- Added mocks for `getRejectedCorpusItems` query.

- Added the first tests that uses those mocks - WIP. Yet to work out how to get the rejected items to load on the page - it seems like our usage of `useEffect` on load is the spanner in the works here. It takes time to work out how to trigger that page load in the tests. I was successful [over there in another tests PR](https://github.com/Pocket/curation-admin-tools/pull/881) with triggering the second mocked query, so it should be achievable here, too - edited to add maybe one day! 🙃.

- Added stubs for tests we should add for this page. 

## Reference

https://getpocket.atlassian.net/browse/BACK-1513


jpetto (approved):
it's a good start!

PR #883, opened 2022-07-18
chore(tests): Remove deprecated warning from Collections tests

nina-py:
## Goal

Remove warnings from unit tests that were polluting the test output - they were especially annoying to wade through in watch mode. The warnings related to a deprecated component property - switched a textarea component from `rows` to `minRows`.

### DIFF
--- a/src/collections/components/CollectionPartnerAssociationForm/CollectionPartnerAssociationForm.test.tsx
+++ b/src/collections/components/CollectionPartnerAssociationForm/CollectionPartnerAssociationForm.test.tsx
@@ -44,7 +44,7 @@ describe('The CollectionPartnerAssociationForm component', () => {
         externalId: '789-xyz',
         name: 'The Cosmos Awaits',
         url: 'https://www.example.com/',
-        imageUrl: 'http://placeimg.com/640/480/people?random=494',
+        imageUrl: 'https://placeimg.com/640/480/people?random=494',
         blurb:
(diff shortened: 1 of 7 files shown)


jpetto (dismissed):
the peace of mind of fewer warnings 🧘

src/collections/components/CollectionPartnerAssociationForm/CollectionPartnerAssociationForm.test.tsx
```diff
@@ -290,9 +290,7 @@ describe('The CollectionPartnerAssociationForm component', () => {
     userEvent.clear(blurbField);
     userEvent.type(
       blurbField,
-      'The sky calls to us vastness is bearable ' +
-        'only through love courage of our questions kindling the energy hidden in ' +
-        "matter the only home we've ever known courage of our questions?"
+      'Hmm, this test tends to time out if this line is too long'
     );
```
nina-py, line 294: Found the reason why some tests randomly time out - it looks like `userEvent.type()` is rather resource-intensive. Here we're testing that the excerpt has a minimum of 10 characters, so a shorter phrase to speed up this test is more appropriate.

jpetto: weird, but okay.

jpetto (approved):

PR #887, opened 2022-07-25
feat(scheduling): [BACK-1543]: Improve the “Scheduled in future” section in Prospecting

nina-py:
## Goal

Improve the “Scheduled in future” section in Prospecting.

## Todos

- [x] Refactor the relevant query out of ProspectingPage component
- [x] Add skeleton layout & connector components
- [x] Add test stubs for new components
- [x] Clean up components to work with real data
- [x] Add unit and integration tests for the new components
- [x] Remove old code (mini schedule card component, grouped list are two I could think of - they are not used anywhere else)
- [x] Tweak the layout to provide new info blocks side-by-side
- [x] Update the ProspectingPage to execute refetch() on changes to the page (another state variable seems to be the most pragmatic option here).
- [x] Abstract away helper functions and add tests
- [x] Reorganise helper functions

## Video walkthrough - see Slack

- https://pocket.slack.com/archives/CTLKW5WCF/p1659344155742449

## References 
Tickets:

- https://getpocket.atlassian.net/browse/BACK-1543


src/curated-corpus/components/ScheduleItemFormConnector/ScheduleItemFormConnector.test.tsx
```diff
@@ -172,13 +172,14 @@ describe('ScheduleItemFormConnector', () => {
     userEvent.click(datePicker);
 
     // Find today's date on the monthly calendar (it's actually a button)
-    const date = screen.getByRole('button', {
+    const dates = screen.getAllByRole('button', {
       name: today.toFormat('d'),
-    }) as HTMLButtonElement;
+    }) as HTMLButtonElement[];
```
Herraj, line 177: I don't think we need to cast it to `HTMLButtonElement` if we fetch them by the button role 🤔

src/curated-corpus/components/ScheduleSummaryCard/ScheduleSummaryCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <TableHead>
+            <TableRow>
+              <TableCell>{headingCopy}</TableCell>
+              <TableCell align="right">#</TableCell>
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {items.map((item) => {
+              return (
+                <React.Fragment key={item.name}>
```
Herraj, line 47: We could replace `<React.Fragment>` with `<>` and add the `key` prop to the `<TableRow>`?

src/curated-corpus/components/ScheduleItemFormConnector/ScheduleItemFormConnector.test.tsx
nina-py: It's often needed, but after whatever happened in `main` that broke this test it's no longer required - I'll remove it 👍.

src/curated-corpus/components/ScheduleSummaryCard/ScheduleSummaryCard.tsx
nina-py: Thanks @Herraj, that was a good catch! Fragment is not needed here at all - I think it was left over after experimenting with other markup (not a table).

nina-py (commented):
I tried to keep refactoring to a minimum here, but several things sneaked through somehow:
- A chunk of functionality + one API call moved to its own component, `ScheduleSummaryConnector` - this part was intentional
- Old code got 🔥! - namely, `ScheduledSurfaceGroupedList` and `MiniScheduleCard` components along with their unit tests
- Helper functions got reorganised a little bit according to their domain. Note that one prospects-related helper function was not moved out of the `helperFunctions.ts` file as it doesn't have unit tests. All but one functions in that file don't have any unit tests - I will submit a ticket to track that shortly.

src/curated-corpus/components/ScheduleSummaryConnector/ScheduleSummaryConnector.tsx
```diff
@@ -0,0 +1,149 @@
...
+      // Prep data for display
+      setPublishers(getGroupedPublisherData(publishers));
+
+      // Extract all topics from scheduled item data
+      const topics =
+        data.getScheduledCorpusItems[0]?.items.map((item) =>
+          getDisplayTopic(item.approvedItem.topic)
+        ) ?? [];
+      // Prep data for display
+      setTopics(getGroupedTopicData(topics));
```
nina-py, line 92: You will see that the two new helper functions, `getGroupedPublisherData` and `getGroupedTopicData`, are very similar. I had persistent thoughts of trying to abstract the code and create one universal function but decided not to remembering @bassrock's guideline to do it if there are three near copies of the same thing in the code, not two 😄.

src/curated-corpus/components/ScheduleSummaryCard/ScheduleSummaryCard.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      { name: 'Publisher 2', count: 1 },
+      { name: 'Another publisher', count: 0 },
+    ];
+    render(<ScheduleSummaryCard headingCopy={'Publishers'} items={items} />);
+
+    // Let's check the last row of the table.
+    // There should be the name of the entry, but not the zero
+    expect(screen.getByText(items[2].name)).toBeInTheDocument();
+    expect(screen.queryByText(items[2].count)).not.toBeInTheDocument();
+  });
```
jpetto, line 37: love this component and tests - specifically how testable they are without integration 😄

src/curated-corpus/helpers/prospects.ts
```diff
@@ -0,0 +1,72 @@
...
+    isTimeSensitive: false,
+    isSyndicated: prospect.isSyndicated ?? false,
+    isCollection: prospect.isCollection ?? false,
+    excerpt: prospect.excerpt ?? '',
+    createdAt: prospect.createdAt ?? 0,
+    createdBy: '',
+    updatedAt: 0,
+    scheduledSurfaceHistory:
+      prospect.approvedCorpusItem?.scheduledSurfaceHistory ?? [],
+  };
```
jpetto, line 71: i like organizing the helpers like this 👍

jpetto (approved):
would like herraj to give this one more review, but looks good to me!

PR #895, opened 2022-07-28
feat(scheduledHistory): [BACK-1488]: Show approver and basic schedule info

Herraj:
## Goal

For existing prospects, show the curator who approved it and the last 3 schedule histories.

Tickets:

- [BACk-1488]

### Screen Capture

https://user-images.githubusercontent.com/16694733/182647400-4a1c634b-2952-48db-87b1-c222324ab42f.mov


src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.tsx
```diff
@@ -85,30 +101,41 @@ export const ExistingProspectCard: React.FC<ExistingProspectCardProps> = (
...
-            )}{' '}
-            <Chip
-              variant="default"
-              color="secondary"
-              label="Already in corpus"
-              icon={<CheckIcon />}
-            />
-          </Box>
+            </Grid>
+          </Grid>
```
Herraj, line 131: Replaced the above jsx from `<Box>` to `<Grid>` to get achieve some spacing. Although, a lot more lines to achieve something little. I could probably use a CSS class and make the styling work with the `<Box>`.

Okay with reverting it if you think it's a bit verbose/lengthy?

nina-py (commented):
Copying here from a Slack conversation on the UI changes:

How do you feel about adding a collapse/more info button in the right hand corner of the prospect card and opening up that schedule history here? I’m worried we’re throwing far too many things on the page, the UI is getting more and more cluttered.

I agree with Jonathan that the schedule history should not take a whole chunk width-wise out of the prospect card, the layout of the main prospect info should remain the same.

I’d also use the user icon rather than the ticked checkbox icon for the “user who added this” (I guess?) info under the image.

Also note that it’s a real possibility that the real estate taken up by the prospect card is going to shrink when https://github.com/Pocket/curation-admin-tools/pull/887 is merged in - instead of taking up 3/4 of the grid it will take up 2/3rds. 

So I think providing the scheduling info but having it hidden by default in a collapsible area would work better than trying to squeeze everything into the available width.

src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.tsx
nina-py: I think it's all right as long as we make corresponding changes to the Prospect card component which also has these pill-style tags.

src/curated-corpus/components/ScheduleHistory/ScheduleHistory.tsx
```diff
@@ -0,0 +1,63 @@
...
+): JSX.Element => {
+  const classes = useStyles();
+  const { data } = props;
+
+  const getScheduledSurfaceName = (surfaceGuid: string) => {
+    return (
+      ScheduledSurfaces.find((surface) => surface.guid === surfaceGuid)?.name ??
+      'Unknown Surface'
+    );
+  };
```
nina-py, line 24: Would this ever happen in practice? Should we throw an error here for Sentry to highlight a data integrity issue? 

Would it be better to show the GUID provided instead of "Unknown Surface" (I am thinking if we have one extra surface added on the backend but the frontend is not updated, it would at least show the code/GUID for that surface instead of "Unknown Surface".

src/curated-corpus/helpers/definitions.ts
```diff
@@ -73,3 +73,13 @@ export type ApprovedItemFromProspect = Omit<
...
+
+// List of scheduled surfaces used to show the readable name by mapping it to it's corresponding guid
+export const ScheduledSurfaces = [
+  { guid: 'NEW_TAB_EN_US', name: 'New Tab (en-US)' },
+  { guid: 'NEW_TAB_DE_DE', name: 'New Tab (de-DE)' },
+  { guid: 'NEW_TAB_EN_GB', name: 'New Tab (en-GB)' },
+  { guid: 'NEW_TAB_EN_INTL', name: 'New Tab (en-INTL)' },
+  { guid: 'POCKET_HITS_EN_US', name: 'Pocket Hits (en-US)' },
+  { guid: 'POCKET_HITS_DE_DE', name: 'Pocket Hits (de-DE)' },
+];
```
nina-py, line 85: I wonder if there's a better way to approach this, even if it takes a bit more time. We managed to get the hardcoded values out of the frontend repo entirely with the addition of the [`getScheduledSurfacesForUser`](https://github.com/Pocket/curated-corpus-api/blob/cefb9fdca562775add3bb37b6c6d3af9c3def117/schema-admin.graphql#L868) query on the backend. 

This returns a list of `ScheduledSurface` objects that contain both the GUID and the name of the scheduled surface, so that it's super easy to generate dropdown options, for example, such as [this bit of code over here](https://github.com/Pocket/curation-admin-tools/blob/cdfbd8c035dfb0711b1696815607a1d35e3a54ab/src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx#L108). 

What if we had another backend query that returned _all_ surfaces, as I imagine we'll need this in more than one place, for example for the upcoming page for each curated item with its own history? 🤔 That would ensure the list of scheduled surfaces is hardcoded in one place only - on the backend.

nina-py: cc @jpetto for their input on this :).

jpetto: i'm having trouble landing confidently on a suggestion, but i think it's this:

for now, hard code the data in this repo, but copy/paste it from our [shared data doc](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2584150049/Pocket+Shared+Data#Source-of-Truth.5).

here's my rationale:

- our shared data is meant to be implemented in a copy/paste manner at this time, and it's already happening in prospect-api and curated-corpus-api, as well as at least one ML repo.
- the long(er)-term plan is to have our shared data available as a service (from _some_ api), which would make a query added to curated-corpus-api unnecessary.
- _if_ we do make a query in curated-corpus-api, someone else is going to use it before long, and then we'll have another client we need to migrate to the actual shared data api when it exists. basically, i am afraid curated-corpus-api would accidentally become the canonical way to retrieve that specific piece of shared data - and maybe it should be! but that decision needs more strategic planning around it.
- we often have to update this repo's generated types when surface shared data is updated anyway - meaning a PR is likely already required for surface updates.

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
```diff
@@ -112,22 +111,26 @@ export const ProspectListCard: React.FC<ProspectListCardProps> = (
...
+              {prospect.isSyndicated && (
+                <Chip
+                  variant="outlined"
+                  color="primary"
+                  label="Syndicated"
+                  icon={<CheckCircleOutlineIcon />}
+                />
+              )}
+            </Grid>
+          </Grid>
```
Herraj, line 133: Updates in this component are swapping out `<Box>` for `<Grid>` just to keep it consistent with the `ExistingProspectCard` component

src/curated-corpus/components/ScheduleHistory/ScheduleHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const classes = useStyles();
+  const { data } = props;
+
+  const [isShowingHistory, setIsShowingHistory] = useState(false);
+
+  const getDisplayDate = (date: string) => {
+    return DateTime.fromFormat(date, 'yyyy-MM-dd')
+      .setLocale('en')
+      .toLocaleString(DateTime.DATE_FULL);
+  };
```
Herraj, line 28: Was thinking about moving this function to the `helperFunctions.ts` file but this function is very specific to this component only. If I were to create a generic `getDisplayDate()` function, it'll have to accommodate for different input types, e.g unix timestamps. So for now I've left it in here.

src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -144,3 +145,23 @@ export const getLocalDateTimeForGuid = (
...
+ * Converts a scheduled surface guid e.g "NEW_TAB_EN_US" to a readable format
+ * @param surfaceGuid
+ * @returns Readable scheduled surface name e.g New Tab (en-US)
+ */
+export const getScheduledSurfaceName = (surfaceGuid: string) => {
+  return (
+    ScheduledSurfaces.find((surface) => surface.guid === surfaceGuid)?.name ??
+    'Unknown Surface'
+  );
+};
```
nina-py, line 167: Let's add tests for these straight away. Happy path + any edge cases (such as the LDAP string being empty/not formatted correctly) should not take too much time for these helper functions.

src/curated-corpus/components/ScheduleHistory/ScheduleHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+  getCuratorNameFromLdap,
+  getScheduledSurfaceName,
+} from '../../helpers/helperFunctions';
+import { Button } from '../../../_shared/components';
+
+interface ScheduleHistory {
+  data: ApprovedCorpusItemScheduledSurfaceHistory[];
+}
+
+export const ScheduleHistory: React.FC<ScheduleHistory> = (
```
nina-py, line 16: It would be great to see comment blocks here for the interface prop and also for the component itself.

src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.test.tsx
```diff
@@ -164,4 +165,39 @@ describe('The ExistingProspectCard component', () => {
...
+        />
+      </MemoryRouter>
+    );
+
+    // the button when clicked shows us the recent scheduled runs
+    const recentScheduledRunsButton = screen.getByRole('button', {
+      name: /view recent scheduled runs/i,
+    });
+
+    expect(recentScheduledRunsButton).toBeInTheDocument();
```
nina-py, line 201: For clarity, it would be good to add a comment here that clicking on the button and making sure we get all data displayed on the page is tested in unit tests for the `ScheduleHistory` component.

nina-py (commented):
A few minor changes and it will be good to go! 👍

src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.tsx
```diff
@@ -85,29 +100,40 @@ export const ExistingProspectCard: React.FC<ExistingProspectCardProps> = (
...
-              color="primary"
-              label={getDisplayTopic(item.topic)}
-              icon={<LabelOutlinedIcon />}
-            />{' '}
-            {item.isSyndicated && (
+          <Grid container spacing={1}>
+            <Grid item>
               <Chip
                 variant="outlined"
                 color="primary"
```
nina-py, line 107: I think some more tweaks are needed here. The markup with the `ProspectCard` component does not appear to be in sync - visually at least. I can see in the video walkthrough that there's less vertical spacing around the tags on the existing prospect card compared to the other card just below it.

I'd also make sure all elements within the card line up to the left - right now the new button doesn't. I'd also capitalise the copy on the new button (we capitalise all other buttons in the app) and change it to to "View recently scheduled runs".

src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.test.tsx
Herraj: Good call, added a top level comment in that test!

src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -144,3 +145,24 @@ export const getLocalDateTimeForGuid = (
...
     .concat(', ', localDateTime.toFormat('t'));
 };
+
+/**
+ * Converts an user's Mozilla ldap username e.g ad|Mozilla-LDAP|jdoe to jdoe
+ * @param ldapString
+ * @returns Curator's username in the following format: flastname
+ */
+export const getCuratorNameFromLdap = (ldapString: string): string => {
+  return ldapString.split('|')[2] ?? 'invalid ldap string';
```
jpetto, line 155: maybe the default case here should just be `ldapString`? that way the actual (weird) value will at least be visible - both for curators to see and us to debug 😄

jpetto (changes requested):
one small request - and i'll defer to nina's notes 😄

src/curated-corpus/helpers/helperFunctions.ts
Herraj: Sure, I was just wondering from a security point of view if we wanted to expose someone's LDAP string but since it's an internal tool, it should be fine.

src/curated-corpus/components/ScheduleHistory/ScheduleHistory.styles.tsx
```diff
@@ -0,0 +1,21 @@
...
+      border: `solid 1px ${theme.palette.secondary.light}`,
+      borderRadius: '5px',
+      width: '100%',
+    },
+    scheduledHistoryWrapper: {
+      margin: '0.5em',
+    },
+    toggleHistoryButton: {
+      paddingLeft: '0px',
+    },
```
Herraj, line 19: @nina-py I've managed to have everything aligned to the left like we had it before I moved stuff from `<Box>` to `<Grid>` except the button itself.

I noticed that the actual button was aligned properly but it was the text inside which had some padding and gave that look of being a bit off from the left. I've removed the `padding-left` for now to fix this but the button does look a bit off when clicked. Open to suggestion 😄

src/curated-corpus/components/ExistingProspectCard/ExistingProspectCard.test.tsx
```diff
@@ -167,6 +167,11 @@ describe('The ExistingProspectCard component', () => {
   });
 
   it('should render schedule history component', async () => {
+    /**
+     * We have separate units tests testing ScheduleHistory component in details in it's corresponding test file
```
nina-py, line 171: Typo - "its", not "it's"

src/curated-corpus/components/ScheduleHistory/ScheduleHistory.styles.tsx
nina-py: I think this is good enough! 👍  If it bugs either of us in the future, we can revisit.

nina-py (approved):
🚀

jpetto (approved):
:shipit:

PR #897, opened 2022-08-02
feat(sidebar): [BACK-1547] Select future dates in “Scheduled in future” section in Prospecting

nina-py:
## Goal

- Added a date picker and a wrapper component for the sidebar - the wrapper holds the sidebar
heading and also the styles that makes it sticky as you scroll down.

- Updated ScheduleSummaryConnector so that it accepts a DateTime (luxon) date format instead of
YYYY-MM-DD.

- Added and updated tests.

- Added small styling/copy tweaks.

## Todos

- [x] Fix newly introduced bugs. Bug number 1: "No results" no longer appears. Fixed, also updated copy to "No stories have been scheduled for this date yet."
- [x] Bug number 2 - scheduling a prospect for the date shown in the sidebar does not update the sidebar.
- [x] Tweak UI - the syndicated split heading needs work, I think - updated to say "X stories, Y syndicated" and removed colour. Increased main heading font size.
- [x] Update and add unit tests - some functionality has been moved around and we have two more components that need tests. Note one test is in `todo` mode - progress on this test is tracked in ticket 1534 (I won't link it here or JIRA will automatically resolve it on merging this PR).

## Video walkthrough

On Slack: https://pocket.slack.com/archives/CTLKW5WCF/p1659524117047119?thread_ts=1659432309.260269&cid=CTLKW5WCF

## Reference

https://getpocket.atlassian.net/browse/BACK-1547


jpetto (approved):
lgtm! will defer to herraj's review though.

src/curated-corpus/components/ScheduleSummaryConnector/ScheduleSummaryConnector.tsx
```diff
@@ -112,19 +108,23 @@ export const ScheduleSummaryConnector: React.FC<
...
       {data && data.getScheduledCorpusItems[0]?.items.length > 0 && (
         <>
           <Typography className={classes.heading} variant="h4">
-            ({data.getScheduledCorpusItems[0]?.syndicatedCount}/
-            {data.getScheduledCorpusItems[0]?.totalCount} syndicated)
+            {data.getScheduledCorpusItems[0]?.totalCount}{' '}
+            {data.getScheduledCorpusItems[0]?.totalCount === 1
+              ? 'story'
+              : 'stories'}
+            , {data.getScheduledCorpusItems[0]?.syndicatedCount} syndicated
```
Herraj, line 127: Should we pull out these conditions into separate variables and put them above the `render()`? I think it would help a bit with readability

src/curated-corpus/components/SidebarWrapper/SidebarWrapper.styles.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      background: 'white',
+      position: 'sticky',
+      top: 0,
+      bottom: 0,
+      paddingTop: '0.5rem',
+      paddingBottom: '0.5rem',
+      zIndex: 5,
```
Herraj, line 15: That's an oddly specific `z-index` 😄 do we have other DOM elements that have a z-index higher than 0 or 1?

Herraj (approved):
Looks good!

src/curated-corpus/components/ScheduleSummaryConnector/ScheduleSummaryConnector.tsx
nina-py: @Herraj, it would help with readability in the component, but the vars would have to go into `onCompleted` as they'd be undefined without data being resolved first, and then TypeScript complains that they need a value anyway before that if they go into the component, so +4 lines of code to improve readability here. I was thinking about a follow-up PR but decided to leave as is for now.

src/curated-corpus/components/SidebarWrapper/SidebarWrapper.styles.tsx
nina-py: I don't think so, not on this page anyway. I think the `5` was lifted straight out of whatever StackOverflow example I was modifying to my use case - traditionally you'd set `z-index` to 100 or 1000 to make sure it's definitely on top of everything else on the page :)

PR #907, opened 2022-08-07
chore(tests): [BACK-1532] Add unit tests for RemoveItemFromScheduledSurfaceModal component

Herraj:
## Goal

- Add an unit test for `RemoveItemFromScheduledSurfaceModal` component
- Abstract the test approved corpus item object and replace the hardcoded ones in existing tests

Tickets:

- [BACK-1532]

[BACK-1532]: https://getpocket.atlassian.net/browse/BACK-1532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/curated-corpus/components/RemoveItemFromScheduledSurfaceModal/RemoveItemFromScheduledSurfaceModal.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+      />
+    );
+
+    // fetching the modal heading and asserting it renders
+    expect(
+      screen.getByText(/remove this item from this scheduled surface/i)
+    ).toBeInTheDocument();
+
+    // fetching the form rendered within this modal by the form's title and asserting it gets rendered
+    expect(screen.getByText(item.approvedItem.title)).toBeInTheDocument();
```
Herraj, line 40: Once again disappointed with react-testing-library. I wanted to fetch the form that is rendered within this modal, using `getByRole()`, which is the correct way to do it. But, the testing library is unable to fetch it. I think it might be nested dom element problem?

In the screenshot below you can see the `<form ...` element is rendered with the correct properties but the testing library fails to fetch it.

Screen Shot 2022-08-07 at 5 35 59 PM

nina-py: Try dropping the `name` attribute on the query. It works for me - the modal is only pulling up one form, so it's not necessary here:

```typescript
const form = screen.getByRole('form');
expect(form).toBeInTheDocument();
```

nina-py (changes requested):
Now that we can get the form, let's add more tests to make sure the functions passed to the modal get executed on pressing form buttons as it's done in other `...Modal` component unit tests.

src/curated-corpus/helpers/approvedItem.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @returns Returns a test object of type ApprovedCorpusItem
+ */
+export const getTestApprovedItem = (
+  options?: Partial<ApprovedCorpusItem>
+): ApprovedCorpusItem => {
+  return {
+    ...approvedCorpusItem,
+    ...options,
+  };
+};
```
nina-py, line 42: 👍 for abstracting this out!

nina-py (approved):

PR #910, opened 2022-08-09
feat(corpus): [BACK-1490] Create a corpus item's full details page

nina-py:
## Goal

Create a brand new page for individual corpus items. This first phase is a "view only" mode, although scaffolding has been added for action buttons to the page.

## Todos

- [x] Add routing for the new page (/curated-corpus/corpus/item/:id/)
- [x] Add a query + generated code to retrieve a corpus item by its
external ID.
- [x] Add scaffolding code for the page and two proposed components (main info + schedule history in the sidebar).
- [x] Flesh out the two layout components
- [x] Add unit tests for the two layout components
- [x] Add a link to each corpus item page from the main Corpus page.
- [ ] Add integration tests for the page - to be done in a follow-up PR to unblock other tickets that need this page to exist.

## Screenshots

Link to item page:

corpus-item-link

First pass on the item page UI (we can refine it later as we add actions, etc.):

curated-item-page

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1490

PR #911, opened 2022-08-09
chore(refactor): split schedule history component into two, split tests as well

Herraj:
## Goal

Split `ScheduleHistory` component and create a `ScheduleHistoryEntries` component that only contains the entries / visuals. The logic and state management stays in `ScheduleHistory` component.


nina-py (approved):
🔥 🍐 
This will allow me to reuse the layout component in #910 for the sidebar, thank you @Herraj!

PR #918, opened 2022-08-16
feat(schedule): [BACK-1556] Link to new Item details page from Schedule page

nina-py:
## Goal

Make the new Corpus Item page accessible both from the Corpus and Schedule pages. The links from the Corpus page were added in BACK-1490, and this ticket follows up with similar changes to the Schedule page.

Due to limited width on the cards, it's hard to fit in another button/link in there well - right now the rightmost button is up against the edge of the card which bothers me (!) 😄. Perhaps we can revisit this at a later point in time and shorten copy on the other buttons or change the "Remove" button from a word to the rubbish bin icon, which would reclaim some space.

- Added a "View" link to corpus item page from the Schedule page.

- Updated unit tests for the relevant component - also simplified
some of those tests by combining separate button rendering tests
into a single test.

- Updated both *CardWrapper unit tests to use `getTestApprovedItem()`, which resulted in some boilerplate code removed.

## Reference

https://getpocket.atlassian.net/browse/BACK-1556

### DIFF
--- a/src/curated-corpus/components/ApprovedItemCardWrapper/ApprovedItemCardWrapper.test.tsx
+++ b/src/curated-corpus/components/ApprovedItemCardWrapper/ApprovedItemCardWrapper.test.tsx
@@ -1,46 +1,12 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
-import {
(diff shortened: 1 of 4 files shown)


jpetto (approved):
lgtm!

but also i agree - it looks cramped 😉. a quick fix may be putting those buttons in a 2x2 table?

PR #919, opened 2022-08-16
feat(schedule): [BACK-1555] Add scheduled date summaries to the Schedule page

nina-py:
## Goal

Add the topic/publisher summary to the schedule page for each date. I initially tried to do this as a collapsible heading (first commit), but it looks fairly ugly - perhaps a drawer sliding out from the side will work better and  not take up space when not needed. 

## Todos

- [x] Figure out the UI - ended up using `Accordion` from MUI 4
- [x] Refactor the `ScheduleSummaryConnector` so that the markup can be used elsewhere without running a self-contained query (as on the Schedule page we'll already have the data ready to pass to the component)

Refactored `ScheduleSummaryConnector` into two components so a second query doesn't need to be run for each date when using the scheduled summary on the Schedule page - now there's a `ScheduleSummaryLayout`
component that is used within the `...Connector` component on the Prospecting page and also directly on the Schedule page with existing data.

- [x] Add/update tests

## Video walkthrough

https://user-images.githubusercontent.com/22447785/185866018-e7ff1a2f-f3c5-4ea5-a397-91b44957700f.mov

## Reference

Tickets:

- https://getpocket.atlassian.net/browse/BACK-1555


src/curated-corpus/integration-test-mocks/getScheduledItems.ts
```diff
@@ -38,7 +38,7 @@ const approvedItem: ApprovedCorpusItem = {
   ],
 };
 
-const scheduledItems: ScheduledCorpusItem[] = [
+export const scheduledItems: ScheduledCorpusItem[] = [
```
nina-py, line 41: Exporting this to use in a unit test

src/curated-corpus/components/ScheduleSummaryConnector/ScheduleSummaryConnector.tsx
```diff
@@ -112,12 +102,12 @@ export const ScheduleSummaryConnector: React.FC<
...
         (data.getScheduledCorpusItems.length < 1 ||
-          data.getScheduledCorpusItems[0]?.items.length < 1) && (
+          scheduledItems.length < 1) && (
           <Typography className={classes.heading} variant="h4">
             No stories have been scheduled for this date yet.
           </Typography>
         )}
-      {data && data.getScheduledCorpusItems[0]?.items.length > 0 && (
+      {data && scheduledItems.length > 0 && (
         <>
```
nina-py, line 111: Managed to get a couple of far-too-long variable names out of the markup with this refactor.

src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -424,9 +433,13 @@ export const SchedulePage: React.FC = (): JSX.Element => {
...
-        <Grid item xs={12}>
-          {!data && <HandleApiResponse loading={loading} error={error} />}
+        {!data && (
+          <Grid item xs={12}>
+            <HandleApiResponse loading={loading} error={error} />
+          </Grid>
+        )}
 
+        <Grid item xs={12}>
           {data &&
```
nina-py, line 443: Small layout tweak to make sure the closing grid tag that contains the `HandleApiResponse` component is right around it instead of somewhere at the bottom of the page.

src/curated-corpus/components/ScheduleSummaryLayout/ScheduleSummaryLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+      (item: { approvedItem: { publisher: string } }) =>
+        item.approvedItem.publisher
+    ) ?? [];
+
+  // Extract all topics from scheduled item data
+  const topics =
+    scheduledItems.map(
+      (item: { approvedItem: { topic: Maybe<string> | undefined } }) =>
+        getDisplayTopic(item.approvedItem.topic)
+    ) ?? [];
```
Herraj, line 40: Potential for abstracting away into helper functions? 😄

src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -438,16 +451,18 @@ export const SchedulePage: React.FC = (): JSX.Element => {
...
+                        <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+                          <Typography className={classes.heading}>
+                            {getDayAndSyndicatedCountHeading(data)}
+                          </Typography>
+                        </AccordionSummary>
+                        <AccordionDetails>
+                          <ScheduleSummaryLayout scheduledItems={data.items} />
+                        </AccordionDetails>
+                      </Accordion>
+                    </Box>
```
Herraj, line 465: Thoughts on making the `<Accordion>` jsx/html part of `<ScheduleSummaryLayout>` jsx? That way we'd have a cleaner jsx here i.e:

```
<Grid item xs={12}>
   <ScheduleSummaryLayout>
</Grid>
```

This is a good option if we plan to always keep this component an accordion component and don't plan to re-use it somewhere else in some other UI format

Herraj (approved):
Good work Nina 🚀 

Last minute thought, what do you think about renaming the `ScheduleSummaryLayout` component to `ScheduleSummaryWrapper` ?

src/curated-corpus/components/ScheduleSummaryLayout/ScheduleSummaryLayout.tsx
nina-py: Yep, but so far these have only been used twice, let's wait for the third time :)

src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
nina-py: I'd have to send the entire query response to it in that case - with the component as is I'm keeping the scope of it smaller and it's easier to unit test with just ScheduledCorpusItem[] sent through.

PR #922, opened 2022-08-17
chore(tests): [BACK-1490] Add integration tests for Curated Item page

nina-py:
## Goal

Make sure the new Curated Item page has integration test coverage

## Todos

- [x] Set up mocks for the query, start on the first test.
- [ ] Figure out why the mock query appears to return `undefined` values and cuts through all TypeScript type checking to fail in helper functions. 
- [ ] Add the remaining tests.


nina-py:
Closing after one last time trying to get these tests running; this needs a fresh look/approach/anything really!

PR #913, opened 2022-08-11
feat(modal): [BACK-1479] Remove duplicate prospect toast and add item details page link to modal

Herraj:
## Goal

We want to replace the duplicate prospect toast message we get when manually adding a prospect. Instead of showing that toast message and the schedule modal, we are going to show the item details page.

## Todos

- [x] Wait for https://github.com/Pocket/curation-admin-tools/pull/910 to be merged and then finalize this component
- [x] Add unit tests once above is completed

## 
Tickets:

- [BACK-1479]

## Implementation Decisions
Created a separate `DuplicateProspectModal` component. It is a very basic and dummy component that doesn't require much internal logic.

[BACK-1479]: https://getpocket.atlassian.net/browse/BACK-1479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Screen Capture
Screen Shot 2022-08-19 at 2 03 14 PM

PR #929, opened 2022-08-22
fix(scheduleItemModal): Refactor boolean condition preventing this modal from rendering

Herraj:
## Goal

Re-open this PR and add fix the bug that was preventing schedule item modal from rendering

## Reference

Original PR that was merged but then reverted: https://github.com/Pocket/curation-admin-tools/pull/913


src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -595,6 +600,14 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
...
       )}
 
+      {approvedItem && (
+        <DuplicateProspectModal
+          approvedItem={approvedItem}
+          isOpen={duplicateProspectModalOpen}
+          toggleModal={toggleDuplicateProspectModal}
+        />
+      )}
+
```
Herraj, line 610: This is the bug fix.  Previously I had:

```
 {approvedItem && (
        <ScheduleItemModal ...  ) && ( <DuplicateProspectModal ... )
```

Which was only rendering the `<DuplicateProspectModal>`. I broke it down into two separate conditionals

src/curated-corpus/components/DuplicateProspectModal/DuplicateProspectModal.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </SnackbarProvider>
+      </MockedProvider>
+    );
+
+    // fetch the modal by its heading and assert it rendered
+    expect(screen.getByText(/duplicate item/i)).toBeInTheDocument();
+
+    // fetch the item title in the modal and assert it rendered
+    // NOTE: I have to include the ":" before the title, otherwise it is unable to fetch
+    expect(screen.getByText(`: ${testApprovedItem.title}`)).toBeInTheDocument();
```
nina-py, line 33: Would using a regex with the title alone work here without the colon included?

src/curated-corpus/components/DuplicateProspectModal/DuplicateProspectModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+   *  opens the link in the new tab and toggles off the modal
+   */
+  const onClick = () => {
+    window.open(
+      `/curated-corpus/corpus/item/${approvedItem.externalId}`,
+      '_blank'
+    );
+
+    toggleModal();
+  };
```
nina-py, line 44: Do we lose state if the link opens in a new tab? I think we do. Is it important? Something to keep in mind but okay to go as is for now, I think.

nina-py (approved):

PR #932, opened 2022-08-23
feat(dismissButton): [BACK-1577] Add dismiss prospect button

Herraj:
## Goal

Add a button to dismiss a prospect.

Tickets:

- [BACK-1577]

### Screen capture

https://user-images.githubusercontent.com/16694733/189983998-c6a202a2-8f3f-42d6-a0a7-2a6145d3b7b1.mov

[BACK-1577]: https://getpocket.atlassian.net/browse/BACK-1577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/curated-corpus/components/actions/DismissProspectAction/DismissProspectAction.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const classes = useStyles();
+  const { prospectId, onDismissProspect } = props;
+
+  const [dismissProspect] = useDismissProspectMutation();
+
+  const onClick = async () => {
+    // call the dismissProspect mutation function and only extract the errors object
+    const { errors } = await dismissProspect({
+      variables: { id: prospectId },
+    });
```
nina-py, line 32: What's the motivation behind only extracting the errors object here?

src/curated-corpus/components/actions/DismissProspectAction/DismissProspectAction.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Note: this test *should* work but doesn't. There are past and open GitHub issues
+   * about it on the Apollo Client project. Unskip this test when the promised overhaul
+   * of the Mocked Provider materialises.
+   *
+   * https://github.com/apollographql/apollo-client/issues/4283 (closed)
+   * https://github.com/apollographql/apollo-client/issues/7167 (open since 2020).
+   *
+   */
+  xit('should render the dismiss button and call the onDismissProspect function with an error message', async () => {
```
nina-py, line 58: This looks familiar - does this issue here, too? Should this comment block simply reference the explanation already elsewhere in the database instead of copying it? 

Also, I'm not sure what the skipped test is meant to assert - could the description be updated so that it's clearer?

nina-py (commented):
Looks good - should there also be a test to make sure the prospect is hidden on dismissal?

src/curated-corpus/components/actions/DismissProspectAction/DismissProspectAction.tsx
Herraj: This function is called with the `prospectId` for the the prospect that has to be dismissed. And that `prospectId` passed in by the parent, so we don't really need to extract the `data` object since we don't really care about any data being returned here.

src/curated-corpus/components/actions/DismissProspectAction/DismissProspectAction.test.tsx
Herraj: Adding a comment inside the test!

Herraj:
> Looks good - should there also be a test to make sure the prospect is hidden on dismissal?

I thought about this test too but unfortunately that would be an integration test for the `ProspectingPage` component since all the prospects are rendered in there.

nina-py (approved):
LGTM! 🚀

jpetto (approved):

PR #939, opened 2022-08-26
feat(corpus item): [BACK-1570] Allow Reject action from corpus item page

nina-py:
## Goal

Let users reject the corpus item right from the Corpus Item page.

## Todos

- [x] Abstract out the functionality used on the Corpus page and reuse it on the Corpus Item page
- [x] Add comments
- [x] Add tests - done, with the exception of sad path - see comments in test file
- [x] Inform the user the corpus item is no longer valid (refetching the data and showing an extended notice now). 
- [x] Rebase branch and squash commits - I'd like to branch off this branch for the other actions while this PR is being reviewed
- [x] Add a video walkthrough
- [x] Add self-review comments

## Video walkthrough

Note that the reason I have to go and add a corpus item before rejecting it is because everything else on Dev has scheduled entries and would not be able to be moved to the rejected pile anyway. The walkthrough shows both the new action off the Curated Item page and also the old one on the Corpus page - this is now powered by same code, just making sure it works after the refactor.

https://user-images.githubusercontent.com/22447785/188348746-e8b28fcd-da5a-424d-b8ba-36159d225549.mov

## Reference

 https://getpocket.atlassian.net/browse/BACK-1570

### DIFF
--- a/src/api/client.ts
+++ b/src/api/client.ts
@@ -7,39 +7,45 @@ import {
   searchCollectionsFieldPolicy,
 } from './field-policies';
 
-const apolloOptions = {
-  cache: new InMemoryCache({
-    typePolicies: {
-      CollectionAuthor: {
-        keyFields: ['externalId'],
-      },
-      CollectionStory: {
-        keyFields: ['externalId'],
-      },
-      Collection: {
-        keyFields: ['externalId'],
-      },
-      ApprovedCuratedCorpusItem: {
-        keyFields: ['externalId'],
-      },
-      RejectedCuratedCorpusItem: {
-        keyFields: ['externalId'],
-      },
-      ScheduledCuratedCorpusItem: {
-        keyFields: ['externalId'],
-      },
-      ScheduledCuratedCorpusItemsResult: {
-        keyFields: ['scheduledDate'],
-      },
-      Query: {
-        fields: {
-          getCollectionAuthors: getCollectionAuthorsFieldPolicy,
-          getCollectionPartners: getCollectionPartnersFieldPolicy,
-          searchCollections: searchCollectionsFieldPolicy,
-        },
+/**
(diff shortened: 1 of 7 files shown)


src/api/client.ts
```diff
@@ -7,39 +7,45 @@ import {
...
-        fields: {
-          getCollectionAuthors: getCollectionAuthorsFieldPolicy,
-          getCollectionPartners: getCollectionPartnersFieldPolicy,
-          searchCollections: searchCollectionsFieldPolicy,
-        },
+/**
+ * Custom type policies for Apollo Cache. Now in their own
+ * variable to feed them into integration tests, too.
+ */
+export const apolloCache = new InMemoryCache({
```
nina-py, line 14: When I was trying to get the last integration test going for this PR, this was identified as one of the possible issues with our integration tests - the fact we haven't previously used the Apollo cache configuration for Mocked Provider.

src/api/client.ts
```diff
@@ -7,39 +7,45 @@ import {
...
+    },
+    ApprovedCorpusItem: {
+      keyFields: ['externalId'],
+    },
+    RejectedCorpusItem: {
+      keyFields: ['externalId'],
+    },
+    ScheduledCorpusItem: {
+      keyFields: ['externalId'],
+    },
```
nina-py, line 33: Along with the refactor into a variable, I updated the type names here - these were pre-refactor names with `Curated` in them and weren't actually having any effect, which may explain the issues with caching we had on the Corpus side of the app.

jpetto (approved):
this looks really good to me. nice refactoring and modulrarization!

Herraj (approved):
Looks good! Just had a thought about approaching this in a different way, i.e creating a custom hook instead of a component in `components/actions`. I'll make a thread in slack!

PR #941, opened 2022-08-29
chore(image): Update generated types and fix collection component pro…

Herraj:
…ps WIP

## Goal

## Todos

- [ ] Outstanding todo
- [x] Completed todo

## Modifying Environment Variables

Omit this section if not applicable.

- [ ] Variables added to SSM
- [ ] Variables added to terraform environment config

## Deployment

Omit this section if not applicable.

Anything we should know about deploying this PR?

## Reference

Documentation here:

- Link to docs

Tickets:

- Link to JIRA tickets

### DIFF
--- a/src/api/fragments/CollectionData.ts
+++ b/src/api/fragments/CollectionData.ts
@@ -1,5 +1,6 @@
 import { gql } from '@apollo/client';
 import { CollectionAuthorData } from './CollectionAuthorData';
+// import { ImageData } from './ImageData';
 
 /**
  * All the properties that are needed to display and edit collections
@@ -14,7 +15,21 @@ export const CollectionData = gql`
     slug
     excerpt
     intro
-    imageUrl
+    image {
+      cachedImages(imageOptions: $imageOptions) {
+        height
+        id
+        url
+        width
+      }
+      caption
+      credit
+      height
+      imageId
+      targetUrl
+      url
+      width
+    }
     language
     status
     authors {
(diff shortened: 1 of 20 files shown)


src/api/fragments/CollectionData.ts
```diff
@@ -14,7 +15,21 @@ export const CollectionData = gql`
...
     excerpt
     intro
-    imageUrl
+    image {
+      cachedImages(imageOptions: $imageOptions) {
+        height
+        id
+        url
+        width
+      }
```
Herraj, line 24: I tried to create two separate fragments --> `ImageData` and `CachedImageData`. Was running into some compilation errors so thought I'd come back to it after making it work intially.

src/collections/components/CollectionListCard/CollectionListCard.tsx
```diff
@@ -21,7 +21,9 @@ interface CollectionListCardProps {
    * An object with everything collection-related in it.
    * Except for stories. We don't need them in the card.
    */
-  collection: Omit<Collection, 'stories'>;
+  collection: Omit<Collection, 'stories' | 'image' | 'partnership'> & {
+    image?: Omit<Image, 'src'> | null;
```
Herraj, line 25: `src` on the new `Type Image` is now deprecated.

I'm not fetching the `src` field under the `image` field on the `Collection` type in the queries/mutations. So I have to omit it from this prop type - won't be necessary once we've replaced all uses of `src` in this code base and can eventually remove the deprecated field from the schema.

Herraj:
Re-doing this work in steps for each separate collections entity.

PR #959, opened 2022-09-14
[BACK-1557] Add unit tests for helper functions

Herraj:
## Goal

Update and add some unit tests for some of the helper functions.

Tickets:

- [BACK-1557]

[BACK-1557]: https://getpocket.atlassian.net/browse/BACK-1557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/curated-corpus/helpers/helperFunctions.ts
+++ b/src/curated-corpus/helpers/helperFunctions.ts
@@ -51,6 +51,7 @@ export const downloadAndUploadApprovedItemImageToS3 = async (
 /**
  * This helper function reads a file, creates an HTML image and
  * assigns the onLoadCallBack function to its onLoad property
+ * NOTE: the optional argument fileReader is only being used for testing purposes as of now
  */
 export const readImageFileFromDisk = (
   file: FileWithPath,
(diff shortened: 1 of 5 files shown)


src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -101,9 +54,11 @@ export const downloadAndUploadApprovedItemImageToS3 = async (
  */
 export const readImageFileFromDisk = (
   file: FileWithPath,
-  onloadCallBack?: VoidFunction
+  onloadCallBack?: VoidFunction,
+  fileReader?: FileReader
```
jpetto, line 58: would be nice to add a comment that this parameter exists for testing purposes. (at least i think that's the case, right?)

jpetto (changes requested):
one tiny request for a comment (usually wouldn't block on a comment request but this is on a function signature), but otherwise looking good! yay tests!

jpetto (approved):
🚢

PR #960, opened 2022-09-29
chores - add husky pre commit hook

Herraj:
## Goal

Add pre-commit hook to catch lint and test errors before CI runs those jobs.

## Implementation Decisions
Do we want to run the tests as well? I think we should in this repo since running just the `npm test` script only tests the newly modified or added tests. This way we can catch any regression


hyperparabolic:
> Do we want to run the tests as well? I think we should in this repo since running just the `npm test` script only tests the newly modified or added tests. This way we can catch any regression

I would support running tests if they do not have external dependencies (unit tests only, no database or other container requirements) and they complete in a reasonable amount of time.

hyperparabolic (approved):
I'm surprised react-scripts still doesn't have a way to execute linting separate from the tests.

Regardless of the choice around running unit tests, this looks fine to me.  Approved.

PR #966, opened 2022-11-14
feat(labels): [BACK-1647] Update `CollectionSearchForm` component

nina-py:
## Goal

So that users can search by one or more labels on the "Search" page.

Did a minor refactor to the search query itself (sending all the filters as one variable instead of individual variables) and updated the "Status" field - the markup on the old one was a bit wonky (labels wouldn't shrink) and it also produced a warning on dev. Additionally, it was missing one of the status values - "Review". Added that, too.

## Video walkthrough

https://user-images.githubusercontent.com/22447785/202342577-e940fa6c-4692-4931-99fa-d2d14538381f.mov

## Reference

https://getpocket.atlassian.net/browse/BACK-1647

### DIFF
--- a/src/collections/components/CollectionSearchForm/CollectionSearchForm.validation.tsx
+++ b/src/collections/components/CollectionSearchForm/CollectionSearchForm.validation.tsx
@@ -4,10 +4,12 @@ import { CollectionStatus } from '../../../api/generatedTypes';
 export const validationSchema = yup.object({
   title: yup.string().min(3),
   author: yup.string().min(3),
+  labels: yup.array(),
   status: yup.mixed<CollectionStatus>().oneOf(Object.values(CollectionStatus)),
-  filterRequired: yup.bool().when(['title', 'author', 'status'], {
-    is: (title: any, author: any, status: any) => !title && !author && !status,
-    then: yup.bool().required('at least one filter is required'),
+  filterRequired: yup.bool().when(['title', 'author', 'status', 'labels'], {
+    is: (title: any, author: any, status: any, labels: any) =>
+      !title && !author && !status && labels.length < 1,
+    then: yup.bool().required('At least one filter is required.'),
     otherwise: yup.bool(),
   }),
 });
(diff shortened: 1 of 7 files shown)


src/collections/components/CollectionSearchForm/CollectionSearchForm.validation.tsx
```diff
@@ -4,10 +4,12 @@ import { CollectionStatus } from '../../../api/generatedTypes';
 export const validationSchema = yup.object({
   title: yup.string().min(3),
   author: yup.string().min(3),
+  labels: yup.array(),
   status: yup.mixed<CollectionStatus>().oneOf(Object.values(CollectionStatus)),
```
nina-py, line 8: This is as far as I went with validation - just an array. https://github.com/Pocket/curation-admin-tools/pull/968 will introduce validation.

src/collections/components/CollectionSearchForm/CollectionSearchForm.validation.tsx
```diff
@@ -4,10 +4,12 @@ import { CollectionStatus } from '../../../api/generatedTypes';
...
-  filterRequired: yup.bool().when(['title', 'author', 'status'], {
-    is: (title: any, author: any, status: any) => !title && !author && !status,
-    then: yup.bool().required('at least one filter is required'),
+  filterRequired: yup.bool().when(['title', 'author', 'status', 'labels'], {
+    is: (title: any, author: any, status: any, labels: any) =>
+      !title && !author && !status && labels.length < 1,
+    then: yup.bool().required('At least one filter is required.'),
     otherwise: yup.bool(),
```
nina-py, line 13: The custom validation error has been updated to include the Labels field - if there's at least one label there, the search will be executed in the absence of other filters. All the old functionality is preserved - e.g. can still search just for authors, etc.

katerinachinnappan: nice!

jpetto (approved):
lgtm

PR #968, opened 2022-11-15
feat(collectionForm): [BACK-1684] Add labels query and validate labels helper function

Herraj:
## Goal

- Validate selected labels against available labels

## Todos

- [x]  Show error message when validation fails under the labels autocomplete field

Tickets:

- [BACK-1684]

[BACK-1684]: https://getpocket.atlassian.net/browse/BACK-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Screen cap 
For the sake of showing an error message in this demo, I've set the validation to not allow any empty labels.

https://user-images.githubusercontent.com/16694733/203424214-ccdc1f66-7e43-43f6-bfdb-0591cf2955b2.mov


src/api/queries/labels.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Get all the labels that are available and can be assigned to a collection.
+ */
+export const labels = gql`
+  query labels {
+    labels {
+      ...labelData
+    }
+  }
+  ${labelData}
+`;
```
nina-py, line 14: We don't really need the query here as the label data is already fetched as part of the `getInitialCollectionFormData` query [here](https://github.com/Pocket/curation-admin-tools/blob/main/src/api/queries/getInitialCollectionFormData.ts) and passed onto the form, and although it will come in handy later when we set up the Labels page, I'd just revert these and all the associated changes here.

src/collections/components/CollectionForm/CollectionForm.tsx
```diff
@@ -141,10 +142,8 @@ export const CollectionForm: React.FC<
     validateOnChange: false,
     validationSchema: getValidationSchema(authorIds),
     onSubmit: (values, formikHelpers) => {
-      // TODO: follow up with a separate PR/ticket that will be solely
-      // responsible for validating the labels before submission
-      // as they don't go through our normal form validation
-      onSubmit(values, formikHelpers, selectedLabels);
+      validateLabels(values.labels, selectedLabels) &&
+        onSubmit(values, formikHelpers, selectedLabels);
```
nina-py, line 146: You would need to pass the `labels` variable here rather than `values.labels` - the `labels` value is the list of all available labels passed to the form component and destructured out of the props [here](https://github.com/Pocket/curation-admin-tools/blob/60824313fb470a7328ca15f1387bfec9f885353a/src/collections/components/CollectionForm/CollectionForm.tsx#L89).

nina-py (commented):
Great start! 👍 

I've found the way we show additional validation errors beneath the form - there's one in the `RejectItemForm` component ([link](https://github.com/Pocket/curation-admin-tools/blob/60824313fb470a7328ca15f1387bfec9f885353a/src/curated-corpus/components/RejectItemForm/RejectItemForm.tsx#L149)). Although you will probably have to set a custom variable to display/hide the error as this validation will not go through Formik.

jpetto (approved):
this looks good to me, but i'll let @nina-py have the final word 😄

nina-py (approved):
LGTM! 🚀

PR #971, opened 2022-11-28
fix(authorName): set minimum author name length to two

Herraj:
## Goal

Update the `AuthorForm` component validation to set minimum author name length to two


Herraj:
I noticed we have the `slug` property required in this form but in the DB (collection-api) we have it as an optional column for the `CollectionAuthor` model. Is there a reason for that?

Screenshot 2022-11-28 at 11 10 25 AM

hyperparabolic (approved):
Looks good to me 👍

jpetto:
> I noticed we have the `slug` property required in this form but in the DB (collection-api) we have it as an optional column for the `CollectionAuthor` model. Is there a reason for that?

great question - unsure. could be worth making an investigation ticket to see if the prod db has any empty values, and if not, potentially make it required in the schema.

PR #973, opened 2022-12-05
Mui 5 rest of authors code

nina-py:
## Goal

Convert the remaining Authors-related pages to MUI 5. Switch theme-related code to MUI 5 - there's a deprecated method that makes our theme object compatible with the new packages.

## Reference

- Ticket TBA


src/collections/components/AuthorInfo/AuthorInfo.tsx
```diff
@@ -20,16 +19,17 @@ interface AuthorInfoProps {
...
     <>
       <Typography
-        className={classes.subtitle}
         variant="subtitle2"
         color="textSecondary"
         component="span"
         align="left"
+        sx={{
+          fontWeight: 400,
+        }}
```
jpetto, line 32: i guess this is nicer/cleaner than having to make a separate file. what happens if/when we need a larger css rule set? would that go into the theme?

src/collections/components/AuthorListCard/AuthorListCard.tsx
```diff
@@ -18,20 +18,33 @@ interface AuthorListCardProps {
...
+        square
+        sx={{
+          margin: '1rem auto',
+          padding: '1.25rem 0.25rem',
+          cursor: 'pointer',
+          '&:active': {
+            backgroundColor: curationPalette.neutral,
+          },
+        }}
+      >
```
jpetto, line 47: i see my previous question is answered here 😄. 

we don't need to be hung up on it during this upgrade, but is the above a good strategy for going forward? these are custom styles - but would they be better made into proper classes and put into the theme?

Herraj: Yeah I think we could benefit from custom css classes here. Although I do understand that might expand the scope a bit.

Herraj: Ideally having custom style variables works the best, that prevents us from hardcoding values. Like we have `curationPalette.neutral` but that requires defining a whole design system and we are not a front end team 😄

src/collections/components/AuthorInfo/AuthorInfo.tsx
nina-py: @jpetto, there's a second styling option for when inline styling is taking up too much space. The JSX node would have to become its own component pretty much, e.g.

```js
const AuthorSubtitle = styled(Component)(() => ({
            fontWeight: 400,
}));
```

and then the `AuthorSubtitle` would have to be used as the pseudo HTML element instead of `Typography` directly. There's also a third-party package option, `tss-react`, which is also worth a look, but I was trying to get away with the minimum of updates here.

PR #976, opened 2022-12-09
feat(mui5): [BACK-1721] Transition Collection pages and components to MUI 5

nina-py:
## Goal

- Updated all components within `src/collections` to MUI 5. This involved a few styling/markup tweaks, updates to props in Autocomplete components used for labels, and deletion of separate styles files with JSS styles.

## Reference

https://getpocket.atlassian.net/browse/BACK-1721


jpetto (approved):
ship!!

PR #982, opened 2022-12-27
feat(mui5): [BACK-1756] Transition Prospecting page to MUI 5

Herraj:
## Goal

Transition Propecting page and the components used within to MUI v5

## Todos

- [x] Surface level mui components
- [x] Custom shared components used within this page component
- [x] Address TODOs, tweaks, clean up
- [x] Verify and fix all unit tests

Tickets:

- [BACK-1756]

[BACK-1756]: https://getpocket.atlassian.net/browse/BACK-1756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Screen cap

https://user-images.githubusercontent.com/16694733/210657048-663e00c2-4745-418b-8c1c-083a2618bcc9.mov


src/curated-corpus/components/AddProspectModal/AddProspectModal.tsx
```diff
@@ -70,7 +70,7 @@ export const AddProspectModal: React.FC<AddProspectModalProps> = (
     setIsManualSubmission,
     setIsRecommendation,
   } = props;
-
+  //TODO: @Herraj save button padding is off, cancel button outline is primary colour instead of neutral
   return (
```
nina-py, line 74: Good catch, I never noticed that! The lighter primary colour for the outline did stand out to me, but I completely forgot it was supposed to be grey instead. The place to fix that would be in the `SharedFormButtons` component (and perhaps even in the custom `Button` component that is used there), not on the Prospecting page. 

Given that it's out of scope for this ticket, would you be happy to file another one specifically for the button issue? I haven't made many changes to those files when I transitioned them over, but we should really take a closer look and probably get rid of the `hollow-outline` etc. types of buttons there. Perhaps we could even go back to default MUI buttons everywhere? We'd do a complete sweep.

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
```diff
@@ -68,12 +75,12 @@ export const ProspectListCard: React.FC<ProspectListCardProps> = (
                 : '/placeholders/collectionSmall.svg'
             }
             alt={prospect.title ?? 'No title supplied'}
-            className={classes.image}
+            sx={{ borderRadius: 4, border: '1px solid lightgray' }}
```
nina-py, line 78: Could we we use one of the app palette colours here instead of the named colour? If using inline styling with the `sx` prop, I've defaulted to using `curationPalette.lightGrey` in cases like this one.

src/curated-corpus/components/SidebarDatePicker/SidebarDatePicker.tsx
```diff
@@ -30,22 +28,27 @@ export const SidebarDatePicker: React.FC<SidebarDatePickerProps> = (
 ): JSX.Element => {
   const { handleDateChange, selectedDate } = props;
 
+  //TODO @Herraj remove datepicker icon and enable datepicker drop down on text field click
   return (
```
nina-py, line 32: We've agreed on Slack that it's not necessary - what is needed here is communicate that change to the curators when we're ready to show them the updated app.

src/curated-corpus/components/SidebarWrapper/SidebarWrapper.tsx
```diff
@@ -3,8 +3,7 @@ import { DateTime } from 'luxon';
 import { useStyles } from './SidebarWrapper.styles';
 import { ScheduleSummaryConnector } from '../ScheduleSummaryConnector/ScheduleSummaryConnector';
 import { SidebarDatePicker } from '../SidebarDatePicker/SidebarDatePicker';
```
nina-py, line 5: Could these imports be cleaned up? Both components should be accessible to import from the index file in the `components` folder, so just from `'../`

src/curated-corpus/components/SplitButton/SplitButton.tsx
```diff
@@ -83,20 +83,23 @@ export const SplitButton: React.FC<SplitButtonProps> = (props) => {
...
+
   return (
     <>
       <ButtonGroup
         variant="text"
-        color="default"
         ref={anchorRef}
         aria-label="split button"
+        sx={{ color: 'rgb(0, 0, 0, 0.87)' }}
       >
```
nina-py, line 95: Could this be consistent with the approaches to specify colours elsewhere? Ideally we'd limit it to just the palette and the gradient colours MUI generates for us out of primary/secondary colours. Perhaps `curationPalette.pocketBlack` would work here?

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -1,9 +1,8 @@
 import React, { useEffect, useState } from 'react';
 import { DateTime } from 'luxon';
-import { Box, Button, Grid, Hidden } from '@material-ui/core';
-import RefreshIcon from '@material-ui/icons/Refresh';
-import FilterListIcon from '@material-ui/icons/FilterList';
-import AddIcon from '@material-ui/icons/Add';
+import { Box, Button, Grid, Hidden } from '@mui/material';
+import { Refresh, FilterList } from '@mui/icons-material';
+import AddIcon from '@mui/icons-material/Add';
```
nina-py, line 5: I'm curious why you used a shorter import here for icons. I'm excited they can be imported from just `@mui/icons-material` without the `Icon` at the end! Yay! But I also like how the code is more readable when it's clear the thing inserted in the markup is an icon.

nina-py: I think the key here is consistency. This is the one and only place I'm seeing these shorter imports for icons. We should do this everywhere or use `XyzIcon` imports everywhere, not both.

nina-py (commented):
Nice work! Small tweaks aside, here's what I think is also within scope for this PR:

- Transitioning to MUI 5 a couple of overlooked components: `AddProspectForm`, `DismissProspectAction`. Apologies if they've already been done but you haven't pushed changes to GitHub yet.

- Fixing any broken tests! Since PRs where the merge target is not `main` don't have them running here. Note that there'll be some console warnings about the new date pickers - apparently it's an issue that's going to be fixed soon by the maintainers of MUI. There are also a couple of warnings from React - complaints about Tabs used on Collections pages - I have not addressed these, but if those mean our tests don't pass then I'll look at fixing those before we merge the feature branch. 

- Note also the comment here: https://github.com/Pocket/curation-admin-tools/pull/984#discussion_r1061160210. I think merging that in should not affect your branch since you've already migrated all the date pickers to the new library.

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
Herraj: Yep I agree. I started with this page and wanted to do a condensed import but turns out not all icons can be imported from the base `/icons-material` so in the other components went with separate imports. Fixed that imports here!

src/_shared/components/TabLink/TabLink.test.tsx
```diff
@@ -35,7 +35,7 @@ describe('The TabLink component', () => {
   it('renders with article count that is larger than 1000', () => {
     render(
       <MemoryRouter>
-        <TabLink count={1235} tabSelected to="/any/path/">
+        <TabLink count={1235} isCurrentTab to="/any/path/">
```
Herraj, line 38: This was giving me an error that `tabSelected` is not a valid prop, turns out they changed the prop name?

src/curated-corpus/components/SidebarDatePicker/SidebarDatePicker.tsx
```diff
@@ -30,22 +28,34 @@ export const SidebarDatePicker: React.FC<SidebarDatePickerProps> = (
 ): JSX.Element => {
   const { handleDateChange, selectedDate } = props;
 
+  // state variable to track date picker open state.
+  // This is being controlled by the onClick on the TextField component within the DatePicker component below
+  const [open, setOpen] = useState(false);
```
Herraj, line 33: @nina-py thoughts on this?

Herraj:
@nina-py As per your request, I've created a [separate chore ticket](https://getpocket.atlassian.net/jira/people/6281ebff56b847006975b76b/boards/98?modal=detail&selectedIssue=BACK-1766) to address the shared form buttons style inconsistencies

src/_shared/components/TabLink/TabLink.test.tsx
nina-py: Nice! 👍

src/curated-corpus/components/SidebarDatePicker/SidebarDatePicker.tsx
nina-py: Nice thing to have, but do we really need it? Two things I can see that create extra maintenance workload here:

1. If we've added this functionality in the sidebar date picker, we should have it everywhere the date pickers are used. And is it worth going and adding extra state variables elsewhere, e.g. on the scheduling form, the scheduling page, anywhere else?
2. If we've added this functionality in the sidebar date picker, we need unit tests for it, too. There are none coming with this PR. And we'd need test coverage for the other places the date pickers are being modified, too.

And so it becomes a question: is it really worth having? I'd stick to default MUI functionality as much as possible. Especially when working with an officially "unstable" component whose API may change in the near future. Getting it to work if something breaks after a library update will be a lot easier if we don't add additional customisations to it.

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -1,9 +1,11 @@
 import React, { useEffect, useState } from 'react';
 import { DateTime } from 'luxon';
-import { Box, Button, Grid, Hidden } from '@material-ui/core';
-import RefreshIcon from '@material-ui/icons/Refresh';
-import FilterListIcon from '@material-ui/icons/FilterList';
-import AddIcon from '@material-ui/icons/Add';
+import { Box, Button, Grid, Hidden } from '@mui/material';
+import Refresh from '@mui/icons-material/Refresh';
+import FilterList from '@mui/icons-material/FilterList';
+import AddIcon from '@mui/icons-material/Add';
```
nina-py, line 6: I see you've gone back to individual imports, but the component names don't have `Icon` in them?

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
```diff
@@ -68,12 +74,15 @@ export const ProspectListCard: React.FC<ProspectListCardProps> = (
...
+            sx={{
+              borderRadius: 4,
+              border: `1px solid ${curationPalette.lightGrey}`,
+            }}
           />
 
           <List dense disablePadding>
             <ListItem disableGutters>
-              <ListItemIcon className={classes.listItemIcon}>
+              <ListItemIcon sx={{ minWidth: '1.5rem' }}>
```
nina-py, line 85: Given that it's used in half a dozen places, would a styled component be worthwhile to add instead? There's already a `StyledListItemAction` on the feature branch, but that one specifies `minWidth` of `2rem`. Could we use that here? Could we update it to `1.5rem`, use it here, and make sure things don't fall apart visually on the Corpus/Schedule pages in the approved card component, too?

nina-py (commented):
Added a few more nitpicky comments, otherwise looking good! 👍

src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
Herraj: Yep, good catch. I've updated the import names!

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
Herraj: Yep good call on re-using components. 

So far I don't really notice a difference when I change that value. I'm not even sure if `minWidth` is doing anything here but testing this out on the corpus page and the prospecting page has made me uncover some other style bugs. Should be quick fixes.

Herraj: I switched the `minWidth` to 1.5rem but didn't take effect. Seems like the style component style is being overriden by the default mui class? I tried to figure out why but will address this in a follow up ticket.

Screenshot 2023-01-06 at 4 51 15 PM

nina-py (approved):

PR #983, opened 2023-01-03
feat(mui5): [BACK-1755] Transition Schedule page to MUI 5

nina-py:
## Goal

Transition the Schedule page and all its components to use MUI 5.

- Updated the SchedulePage component and almost all other components used on this page to use MUI 5 (SplitButton has had updates made in #982, so I have not updated it here).

- Made a few visual tweaks - DatePicker fields are now slim (with smaller padding around the data) so that their vertical width now matches input fields on forms and buttons.

- Updated page container width to use a default MUI value for width ("xl") instead of a fixed pixel width.

- Fixed a couple of `Header` component unit tests - I'm sure I've fixed them in a previous PR, perhaps did not commit?

## Deployment

Note this PR's merge target should be `feature/mui-5-upgrade`, not `main`.

## Screenshots

Note the formatting in the corpus item card is slightly off in the first screenshot (notably, the title is no longer bold) - this because the page uses MUI 5 and that card component is still on MUI 4, as it's also used on the Corpus page and updating it will likely be the bulk of updates of that page):

schedule-page

And here's the "Reschedule Item" modal with new, slim date picker field:

reschedule-modal

## Reference

https://getpocket.atlassian.net/browse/BACK-1755


src/_shared/components/FloatingActionButton/FloatingActionButton.tsx
```diff
@@ -24,14 +24,12 @@ export const FloatingActionButton: React.FC<FloatingActionButtonProps> = (
...
-          right: '1.5rem',
-          bottom: '1.5rem',
-          left: 'auto',
-        },
+        position: 'fixed',
+        margin: '0rem',
+        top: 'auto',
+        right: '1.5rem',
+        bottom: '1.5rem',
+        left: 'auto',
```
nina-py, line 32: I don't know how this dumb mistake snuck in 😊 , but I've fixed it now.

src/_shared/pages/LandingPage/LandingPage.tsx
```diff
@@ -16,7 +16,7 @@ export const LandingPage = (): JSX.Element => {
         productLink=""
         menuLinks={[]}
       />
-      <StyledContainer>
+      <StyledContainer maxWidth="xl" disableGutters>
```
nina-py, line 19: Rather than having our own custom `maxWidth` CSS style set to 1260 px, let's use the nearest MUI default.

src/curated-corpus/components/ScheduleItemFormConnector/ScheduleItemFormConnector.tsx
```diff
@@ -102,10 +102,7 @@ export const ScheduleItemFormConnector: React.FC<
     });
 
   // What to do when the user clicks on a date in the calendar.
-  const handleDateChange = (
-    date: MaterialUiPickersDate,
-    value?: string | null | undefined
-  ) => {
+  const handleDateChange = (date: any, value?: string | null | undefined) => {
```
nina-py, line 105: I don't like setting the `date` var to `any` here and in the other part of the code that's moved to the new date pickers but the date picker upgrade has been troublesome enough! 🙃

src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -367,64 +367,71 @@ export const SchedulePage: React.FC = (): JSX.Element => {
...
-                      endDate
-                    );
-                  }}
-                >
-                  Get schedule
-                </Button>
+            <LocalizationProvider dateAdapter={AdapterLuxon}>
+              <Grid container spacing={2} alignItems="center">
+                <Grid item sm>
+                  <DatePicker
```
nina-py, line 378: `sm` had to be added to the Grid item component here so that the two date fields line up with the button again.

Also, it would be nice to use a `DateRangePicker` instead of two `DatePicker` components here, but both are coming from MUI X, which are still marked as "unstable" in the list of components: https://mui.com/material-ui/discover-more/roadmap/#new-components, so it's a task for another day.

PR #987, opened 2023-01-09
feat(mui5): [BACK-1769] Transition Labels page to MUI 5

nina-py:
## Goal

Make sure recent updates to the frontend app have also been migrated to MUI 5. 

- Updated LabelListPage and LabelListCard components to work with MUI 5.

## Deployment

Note this is to be merged to the feature branch (`feature/mui-5-upgrade`), not `main`. 

## Reference

https://getpocket.atlassian.net/browse/BACK-1769


src/collections/components/LabelListCard/LabelListCard.styles.tsx
```diff
@@ -1,28 +0,0 @@
...
-    },
-    title: {
-      fontSize: '1.25rem',
-      fontWeight: 500,
-    },
-    [theme.breakpoints.down('sm')]: {
-      title: {
-        fontSize: '1rem',
-      },
-    },
```
nina-py, line 26: I dropped this style tweak from the styles that were carried over to the MUI5-compatible syntax as it would require a more thorough update than just throwing everything into an `sx` property. Given that we don't know how this page is going to look like once development of those features resumes, I think this is justified.

Herraj (approved):

PR #998, opened 2023-01-12
INFRA-900: Add createLabel mutation and Create label form & modal

katerinachinnappan:
## Goal

* Add a `createLabel` mutation & generate API types
    * Create `AddLabelForm` component to render create label form
    * Create `AddLabelModal` to render the `add label` modal from `LabelListPage`
    * Update `LabelListPage` to render `AddLabelModal`
* Add tests

## Todos

- [x] Test in dev when reviewed

## Tickets:

- [https://getpocket.atlassian.net/browse/INFRA-900](https://getpocket.atlassian.net/browse/INFRA-900)

## Demo

https://user-images.githubusercontent.com/16909783/212435520-922a0784-0394-4bf8-90d1-564cecc33ec8.mov


src/collections/components/AddLabelForm/AddLabelForm.validation.tsx
```diff
@@ -0,0 +1,9 @@
+import * as yup from 'yup';
```
katerinachinnappan, line 1: Minimal validation for now to move this PR. More validation to come in a separate PR after establishing what else needs to be validated.

nina-py: I think it's a good idea to implement validation right from the start. We know that labels should be slug-compatible, so reusing validation rules in place for Collection/CollectionStory/CollectionAuthor slugs should be fine here. Also need to enforce this on the backend - I assume you have this under control already.

katerinachinnappan: @nina-py yep, that is a good point however we still need to collect requirements like min/max size and if we allow non-slug style. Also, for example if it is a `region` kind of label, we might want to validate it starts with `region`. We don't have all the requirements for now and set aside for another ticket.

src/collections/components/AddLabelModal/AddLabelModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+  props
+): JSX.Element => {
+  const { isOpen, toggleModal, setLabelsList } = props;
+  const [isLoaderShowing, setIsLoaderShowing] = useState<boolean>(false);
+  const [createLabelMutation] = useCreateLabelMutation();
+  // Prepare state vars and helper methods for API notifications
+  const { showNotification } = useNotifications();
+
+  const onSubmit = (
+    values: FormikValues,
```
nina-py, line 39: There's quite a lot happening in this component. It's good that the API call is not being made straight from the Labels page as some of our older code does, but the other `XyzModal` components are not set up in the same way. They're solely responsible for Modal markup and managing their own state (e.g., open/closed). 

Then each either calls the form directly (older pattern) or call a Connector component (current pattern) which has next to no markup - it just calls the form and passes on the `onSubmit` function - but encapsulates all the business logic of calling the API and getting the response. Let's follow this pattern here. Then it will be relatively easy to set up unit tests for the UI-only components and integration tests for the component responsible for calling the API. Have a look at the flow from `ScheduleItemModal`, it's probably the most complete example.

src/collections/components/AddLabelModal/AddLabelModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Toggle the AddLabelModalProps to show/hide as necessary.
+   */
+  toggleModal: VoidFunction;
+
+  /**
+   * state variable from Parent component LabelList Page
+   * */
+  setLabelsList: React.Dispatch<React.SetStateAction<Label[]>>;
+}
```
Herraj, line 22: We could create a function that calls the state setter function instead of passing the state function directly to the child

```suggestion
  setLabelsList: (labelsList: Label[]) => void;
}
```

And then we can pass a function to the child component as a prop which calls the state setter e.g

```
// something like this
const setLabels = (labelList: Label[]) => {setLabelList(labeList);}
```
Apologies for misleading you!

nina-py: What about not manipulating the contents of the list directly? I know our cache management (specifically Apollo cache) within the app is a hot stinking mess, but we should at least make it a _consistent_ mess :). Then, once we've got some time set aside to deal with it, we could refactor it in a single sweep. 

@katerinachinnappan , what we'd usually do here is flush the cache for the `labels` query on the page by destructuring a helper function - `refetch` - from the original query and using it to flush the cache once the mutation is completed. It is supposed to work without this manual intervention but doesn't. [Example here](https://github.com/Pocket/curation-admin-tools/blob/9255721fb0cd1ba3d1ad1e5721defa2741085318/src/curated-corpus/pages/CorpusPage/CorpusPage.tsx#L25).

src/collections/components/AddLabelFormConnector/AddLabelFormConnector.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * state variable for list of lables
+   * */
+  setLabelsList: React.Dispatch<React.SetStateAction<Label[]>>;
+}
+
+/**
+ * Parent component for the AddLabelFormConnector component
+ */
+export const AddLabelFormConnector: React.FC<AddLabelFormConnectorProps> = (
```
nina-py, line 22: Did you mean to say in the comment "for the AddLabelForm" here?

src/collections/components/AddLabelFormConnector/AddLabelFormConnector.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+  createLabel2SuccessMock,
+  createDuplicateLabelErrorMock,
+  createMinCharLabelErrorMock,
+  createBadCharLabelErrorMock,
+} from '../../integration-test-mocks/createLabels';
+
+describe('AddLabelFormConnector', () => {
+  let mocks = [];
+  const toggleModal = jest.fn();
+
```
nina-py, line 18: Truly excellent integration test coverage! 🥇

nina-py (commented):
Looking great! The tests are especially appreciated.

I'd make changes to make list state management in line with how it's done elsewhere in the codebase, and then it'll be good to go.

src/collections/components/AddLabelModal/AddLabelModal.tsx
Herraj: Yeah even better^ lets use `refetch`, that way we get the label list from the query itself and maintain that one source of truth

src/collections/components/AddLabelFormConnector/AddLabelFormConnector.tsx
```diff
@@ -0,0 +1,81 @@
...
+        }
+      })
+      .catch((error: Error) => {
+        setIsLoaderShowing(false);
+        showNotification(error.message, 'error');
+      });
+    setIsLoaderShowing(true);
+
+    formikHelpers.resetForm();
+  };
```
Herraj, line 73: If you wanted a clean this up a little, we have a nice `useRunMutation` hook https://github.com/Pocket/curation-admin-tools/blob/main/src/_shared/hooks/useRunMutation/useRunMutation.ts

It performs the same logic as your `onSubmit()` here!

src/collections/components/AddLabelFormConnector/AddLabelFormConnector.tsx
```diff
@@ -0,0 +1,66 @@
...
+      { variables: { name: values.labelName } },
+      `"${values.labelName} has been created"`,
+      () => {
+        setIsLoaderShowing(false);
+        toggleModal();
+        formikHelpers.setSubmitting(false);
+
+        if (refetch) {
+          refetch();
+        }
```
Herraj, line 50: I think we can remove this and just pass the `refetch` callback as the last parameter for `runMutation()` call. That exact same logic exists in the `runMutation()`

katerinachinnappan: ah good catch, missed that

Herraj (approved):
:shipit:

nina-py (approved):
🚀

PR #1070, opened 2023-03-29
OSL-168: Adding hide list functionality

katerinachinnappan:
## Goal

* Created a reusable (for future purposes) `ShareableListModal` and related components for the moderation form which currently hides a list. 
* Added integration tests.

## Todos

- [ ] deploy to dev

## Link to JIRA tickets

- [https://getpocket.atlassian.net/browse/OSL-168](https://getpocket.atlassian.net/browse/OSL-168)

## Demo

https://user-images.githubusercontent.com/16909783/228951560-8987c86d-0eca-40c4-8996-5537a5a0d86d.mov


nina-py (commented):
Great job! A couple of minor comments:

- Do we really need to say "This list is hidden" if the tags/pills in line with the list item already show if the list is visible or hidden (the secondary colour, red brick ones?)
- Would it be better to keep the "Hide this list" button but make it greyed out/disabled in case if the list has already been taken down?
- The optional "Moderation reason" field. Could we make it a multiline text area? It's possible takedown comments will include some details - e.g. reference to discussions with the legal team, so a textarea field that can be resized might serve us better here.

nina-py (approved):
Great job! 🚀

PR #1097, opened 2023-05-05
feat(prospecting): [INFRA-1006, INFRA-1007] Filter by publisher

nina-py:
## Goal

Allow curators to include or exclude certain publishers from the list of prospects on the Prospecting page (see video walkthrough below).

- Added a new component that contains additional UI elementsfor filtering by publisher.

- Added filtering by publisher (both include and exclude) on the frontend, with prospects already pre-loaded on the page.

- Added tests for the new component.

- Updated a couple of existing components - removed inconsistencies in presentation of data.

## Video Walkthrough

https://user-images.githubusercontent.com/22447785/237024030-5dfe6623-aa24-4f46-b6f6-09a8af435649.mov

## Implementation Decisions

After several attempts at getting things right, I implemented this filter entirely on the frontend. There was a recent PR that allowed filtering prospects on the backend but - (1) it returned a subset of results as the filters were applied post-fetching results, and (2) it was a jarring experience having the entire prospect set refresh instead of having some publishers filtered out, with or without the confirmation modal. I think the entire set of filters on the page needs to be redesigned and redeveloped into a single form, but that is a task for another day.

Additionally, the Prospecting page has so much going on, and its refactor is probably some time away from becoming a reality, that I decided a light touch was needed - an absolute minimum of changes to get this in.

## Reference

- https://getpocket.atlassian.net/browse/INFRA-1006
- https://getpocket.atlassian.net/browse/INFRA-1007


src/curated-corpus/pages/ProspectingPage/ProspectingPage.tsx
```diff
@@ -692,20 +705,50 @@ export const ProspectingPage: React.FC = (): JSX.Element => {
...
+                else if (
+                  !excludePublisherSwitch &&
+                  !prospect.publisher
+                    .toLowerCase()
+                    .includes(filterByPublisher.toLowerCase())
+                ) {
+                  return;
+                }
+              }
+
```
Herraj, line 749: We could extract the above logic into a separate helper function but I guess we'll have to do that as part of the larger refactor for this component that we are waiting for 😄

Herraj (approved):
🚀

PR #1116, opened 2023-06-15
[TML-75] Add title fix button

vazish:
## Goal
Based on meeting we had with Editorial team, we wanted to add a button that fixes the title and the straight apostrophe's in titles when populating from a url.

Created two helper functions for these two fixes:

Add curly quotes and fix AP style title case for titles in collections (see tests also)

Per request from Editorial team, added a 'fix excerpt' button as well

## Reference

Documentation here:
1. Casing follows AP style here - https://headlinecapitalization.com/
2. Regular single and double quotes replaced by curly ones

Tickets:

[Jira](https://getpocket.atlassian.net/jira/software/projects/TML/boards/123?assignee=637ba9dca593cb822e93c33f)
Screenshot 2023-06-15 at 12 27 14

### DIFF
--- a/src/_shared/utils/applyApTitleCase.ts
+++ b/src/_shared/utils/applyApTitleCase.ts
@@ -1,0 +1,47 @@
+export const STOP_WORDS =
+  'a an and at but by for in nor of on or so the to up yet';
+
+export const SEPARATORS = /(\s+|[-‑–—,:;!?()])/;
+
+/**
+ * Capitalize first character for string
+ *
+ * @param {string} value
+ * @returns {string}
+ */
+const capitalize = (value: string) => {
+  if (!value) {
+    return '';
+  }
+  return value.charAt(0).toUpperCase() + value.slice(1);
+};
+
+/**
+ * Helper to convert text to AP title case
+ * adapted from https://github.com/words/ap-style-title-case
+ * text should match https://headlinecapitalization.com/
+ *
+ * @param {string} [value]
+ * @returns {string}
+ */
+export const applyApTitleCase = (value: string): string => {
+  if (!value) {
+    return '';
+  }
+  // split by separators, check if word is first or last
+  // or not blacklisted, then capitalize
+  const stop = STOP_WORDS.split(' ');
+  return value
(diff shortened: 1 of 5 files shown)


Herraj (dismissed):

jsrawan-mobo:
Was it just the title and not the excerpt as well ? 

When they add a story - did we discuss just fixing it live versus  the button?

src/_shared/utils/applyApTitleCase.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param {string} [value]
+ * @returns {string}
+ */
+export const applyApTitleCase = (value: string): string => {
+  if (!value) {
+    return '';
+  }
+  // split by separators, check if word is first or last
+  // or not blacklisted, then capitalize
+  const stop = STOP_WORDS.split(' ');
```
jsrawan-mobo, line 33: seems silly to split a const every time - shouldn't you do that once at the top?

vazish:
> Was it just the title and not the excerpt as well ?
> Title only. Excerpt would follow different rules from the headline ones

> When they add a story - did we discuss just fixing it live versus the button?
I talked to Amber. We decided to use a button so the Editors could see the change live instead of it happening in the background.

src/_shared/utils/applyApTitleCase.ts
vazish: Ok I updated, pls approve, can't merge otherwise

src/_shared/utils/applyCurlyQuotes.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Helper to replace opening and closing curly single and double quotes
+ *
+ * @param text
+ * @returns {string}
+ */
+
+export const applyCurlyQuotes = (text: string): string => {
+  if (!text) return '';
+  return text
+    .replace(/'\b/g, '\u2018') // Opening singles
```
rolf-moz, line 11: You might want to define quote constants for future tests

rolf-moz (dismissed):

PR #1133, opened 2023-07-11
TML-86 show more excerpt options in creating collection

rolf-moz:
## Goal
For https://getpocket.atlassian.net/browse/TML-86 we found that the OpenGraph description (which typically is article subhead) was superior to the Parser description (which it typically first few sentences of the article).

We provide both descriptions when available and allow switching between the two. The OG description is prioritized.

## Todos

- [ ] add more tests

Yes

## Deployment
Companion TML-86 PR for corpus-api should be merged first
Confirm shared list API rollback is merged as well 

## Implementation Decisions
We are using a simple backend call to get OG description
Ideally it would be best added to parser: https://getpocket.atlassian.net/wiki/spaces/PE/pages/2988441601/Spec+Add+article+description+to+Item


src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -121,78 +163,96 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
-            imageUrl = data.getItemByUrl.images[0].src!;
+    onCompleted: ({ getOpenGraphFields }) => {
+      if (getOpenGraphFields?.description) {
+        const excerpt = getOpenGraphFields.description;
+        formik.setFieldValue('excerpt', excerpt);
+        setOgExcerptText(excerpt);
+      }
+    },
+    onError: () => {
+      // Errors are silent as this is an optional service
```
Herraj, line 176:
```suggestion
      return
```

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -121,78 +163,96 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+        // If this is used to add a story and only the URL is visible,
+        // show the other fields now that they contain something
+        // even if the parser can't process the URL at all.
+        setShowOtherFields(true);
+      },
+      onError: (error: ApolloError) => {
+        // Show any other errors, i.e. cannot reach the API, etc.
+        showNotification(error.message, 'error');
+      },
+    });
```
Herraj, line 255: Looks like line 180 - 255 is just the previous existing code but got moved down a bit?

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -325,19 +408,35 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+                      ogExcerptText,
+                      DescriptionTextStates.CanInsertParserDesc
+                    )}
+                </Grid>
               </MarkdownPreview>
             </Grid>
             <Grid item xs={12}>
               <Button buttonType="hollow" onClick={fixExcerpt}>
-                Fix excerpt
+                Fix quotes
```
Herraj, line 439: Was this intentional? onClick is pointing to the excerpt function 🤔 haven't looked at the UI so I could be wrong here.

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -227,6 +292,24 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+        onClick={() => {
+          formik.setFieldValue('excerpt', textToInsert);
+          setDescriptionState(newDescriptionState);
+        }}
+        sx={{ textDecoration: 'none' }}
+      >
+        {linkDescriptionText}
+      </StyledLink>
+    </Tooltip>
+  );
```
Herraj, line 311: I would move this into a separate helper / shared component since it's returning jsx and we try to keep just one jsx component per file.

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -82,6 +102,28 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+  const [ogExcerptText, setOgExcerptText] = useState<string>('');
+  const [parserItemExcerptText, setParserItemExcerptText] =
+    useState<string>('');
+  const [descriptionState, setDescriptionState] =
+    useState<DescriptionTextStates>(
+      DescriptionTextStates.IncompleteDescription
+    );
+
+  useEffect(() => {
+    console.log(`OG: ${ogExcerptText} Parser: ${parserItemExcerptText}`);
```
Herraj, line 115: Was meant to be taken out?

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -82,6 +102,28 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+    if (ogExcerptText && parserItemExcerptText) {
+      if (ogExcerptText === parserItemExcerptText) {
+        setDescriptionState(DescriptionTextStates.MatchingDescriptions);
+      } else {
+        setDescriptionState(DescriptionTextStates.CanInsertParserDesc);
+      }
+    } else {
+      setDescriptionState(DescriptionTextStates.IncompleteDescription);
+    }
```
Herraj, line 124: Thoughts on extracting into a helper function?

src/collections/components/StoryForm/StoryForm.tsx
```diff
@@ -82,6 +102,28 @@ export const StoryForm: React.FC<StoryFormProps & SharedFormButtonsProps> = (
...
+  }, [ogExcerptText, parserItemExcerptText]);
```
Herraj, line 125: I am wondering if we really need a `useEffect()` here? We try to limit the use of this hook and try to use it where we absolutely need it. Do you think we just have this logic (line 116 - 124) in a just a regular function, given we will always have access to the state variables in this hook's dependency array? 

Just a thought!

src/collections/components/StoryForm/StoryForm.tsx
rolf-moz: the code only fixes the quotes so I thought it would be more clear. maybe it is confusing?

src/collections/components/StoryForm/StoryForm.tsx
rolf-moz: Description state can be changed from user interaction OR from new data coming in, so it is really needs this treatment. The code could be simpler if I could get the formik text field data, though I think the users might not like some of the behavior quirks. Let me know if you want to discuss further

src/collections/components/StoryForm/StoryForm.tsx
rolf-moz: ok, I split it up

src/collections/components/StoryForm/StoryForm.tsx
Herraj: Sounds good and I think it does make sense here given that new data could come in.

src/collections/components/TextSwitchLink/TextSwitchLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+      textToInsert = parserItemExcerptText;
+      break;
+    case DescriptionTextStates.CanRestoreOGDesc:
+      uiText = 'Restore OpenGraph Description';
+      nextDescriptionState = DescriptionTextStates.CanInsertParserDesc;
+      textToInsert = ogExcerptText;
+      break;
+    default:
+      return null;
+  }
```
Herraj, line 71: Nice!

Herraj (approved):

PR #1140, opened 2023-07-20
Fix quotes to collection excerpt & intro

vazish:
## Goal

Add fix quotes buttons to excerpts & intro part of collection
Fix tests

Tickets:

- [Jira](https://getpocket.atlassian.net/browse/TML-127)

### DIFF
--- a/src/collections/components/CollectionForm/CollectionForm.tsx
+++ b/src/collections/components/CollectionForm/CollectionForm.tsx
@@ -31,6 +31,7 @@ import {
   Label,
 } from '../../../api/generatedTypes';
 import { applyCurlyQuotes } from '../../../_shared/utils/applyCurlyQuotes';
+import { applyApTitleCase } from '../../../_shared/utils/applyApTitleCase';
 
 interface CollectionFormProps {
   /**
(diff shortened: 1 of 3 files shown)


src/collections/components/CollectionForm/CollectionForm.tsx
```diff
@@ -199,12 +208,21 @@ export const CollectionForm: React.FC<
...
       <Grid container spacing={3}>
         <Grid item xs={12}>
-          <FormikTextField
-            id="title"
-            label="Title"
-            fieldProps={formik.getFieldProps('title')}
-            fieldMeta={formik.getFieldMeta('title')}
-          />
+          <Box display="flex">
+            <Box flexGrow={1} alignSelf="center" textOverflow="ellipsis">
```
rolf-moz, line 212: Do you think it might simplify all these cases to create a component that includes FormikTextField internally?
<FormicCorpusTextField id="title" label="Title" centering_type="box" formic={formic}/> or something like that?

rolf-moz (approved):

src/collections/components/CollectionForm/CollectionForm.tsx
vazish: Yes, it might be tricky with the css though. Let me merge this for now and do a refactor after.

PR #1148, opened 2023-11-30
Mc 269 add toggle to use og excerpt for corpus item

Herraj:
## Goal

## Todos

- [ ] Outstanding todo
- [x] Completed todo

## Modifying Environment Variables

Omit this section if not applicable.

- [ ] Variables added to SSM
- [ ] Variables added to terraform environment config

## Deployment

Omit this section if not applicable.

Anything we should know about deploying this PR?

## Reference

Documentation here:

- Link to docs

Tickets:

- Link to JIRA tickets


src/curated-corpus/components/ApprovedItemForm/ApprovedItemForm.tsx
```diff
@@ -106,6 +117,34 @@ export const ApprovedItemForm: React.FC<
     formik.setFieldValue('imageUrl', url);
   };
 
+  // state variable to store and set Open Graph excerpt
+  const [ogExcerpt, setOgExcerpt] = useState<string>('');
+
+  /**
+   * This calls the query on component mount. This isn't ideal because the editors might not even want the OG excerpt so we
```
efixler, line 124: If the web page is end-user facing 💯 . In this case -- where the usage of this page is low and only for curators -- it doesn't read to me as a major problem. (If anything it means that the editors can toggle and choose and that'll be performant, right?)

efixler:
What's the url for dev to test this?

Herraj:
> So, I tested the above -- the Save does work to save the OG description.
> 
> When I reopen the Corpus Item I see the following:
> image
> 
> Is that actually true? We didn't update the parser info to match the OG Info, did we?

I think you've found a bug! I think I know why this is happening. The excerpt comparison is done on `approvedItem.excerpt` (parser excerpt before saving) and `ogExcerpt`. After saving the og excerpt, `approvedItem.excerpt` is now og excerpt. So that's why it's showing us that. I can look into this.

katerinachinnappan (approved):

PR #1156, opened 2024-01-30
Add SlateScheduler_V2 for GCP

jsrawan-mobo:
## Goal

- Added prospect SlateScheduler_V2
- Catch case where there parser hydration fails (usually on local dev)

## Link to JIRA tickets
https://mozilla-hub.atlassian.net/browse/MC-541


src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
```diff
@@ -69,8 +69,13 @@ export const ProspectListCard: React.FC<ProspectListCardProps> = (
...
 
-  const timeToRead = parserItem.timeToRead ? (
-    `${parserItem.timeToRead} min(s)`
+  let _parserItem = parserItem;
+  if (!parserItem) {
+    _parserItem = {};
+  }
+
+  const timeToRead = _parserItem.timeToRead ? (
+    `${_parserItem.timeToRead} min(s)`
```
Herraj, line 78: Can you please explain what lines 72-75 are intended for? 

I think lines 77-81 here are essentially doing the same thing as what we had before 🤔

jsrawan-mobo: if  `_parserItem`  - this causes the UI to crash on line 77.  Oh we can add a ? there

Herraj: Was it causing issues how we had it before? 

If `parserItem` object is `null/undefined` and/or `timeToRead` property on that object is `null/undefined/falsey` (it will be if the `parserItem` itself is falsey), then the ternary expression would default to the clause after the ` : `. 

So line `74` doesn't really do anything here 🤔 unless I'm missing something

src/api/generatedTypes.ts
```diff
@@ -1353,6 +1353,7 @@ export enum ProspectType {
   RssLogistic = 'RSS_LOGISTIC',
   RssLogisticRecent = 'RSS_LOGISTIC_RECENT',
   SlateScheduler = 'SLATE_SCHEDULER',
+  SlateScheduler_V2 = 'SLATE_SCHEDULER_V2',
```
efixler, line 1356: Why are we making a new feed?

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
katerinachinnappan: @Herraj @jsrawan-mobo is this for when running the tool locally (along with prospect-api + admin-api locally) & the parserItem is null?

src/api/generatedTypes.ts
jsrawan-mobo: We want to validate the GCP vs AWS side by side.   The editors don't need to use this feed until we give finish this spike.

src/curated-corpus/components/ProspectListCard/ProspectListCard.tsx
jsrawan-mobo: Yes this happens locally all the time.  It could happen in production also.

if parserItem is undefined,  line 77 throws because you can't dereference.  I added a test to illustrace.  Lines 137 later on had the right check.

jsrawan-mobo: > So line `74` doesn't really do anything here 🤔 unless I'm missing something

you can't  do `undefined.attr` or `null.attr` you can do `false.attr` 
I think your meant chaining

`undefined' && undefined.attr1 && undefined.attr1.attr2`

is same as  `undefined?.attr1?.attr2`

Herraj (approved):

Herraj (approved):

katerinachinnappan (approved):

PR #1165, opened 2024-02-19
feat(schedule): [MC-572] Color-code 'last scheduled' in red for recent occurences, same surface only

nina-py:
## Goal

Color-code 'last scheduled' in red for recent occurrences, same surface only. Rather than color-coding the recently scheduled items across all surfaces, only highlight them if they've recently appeared on the same surface only.

## Video walkthrough

https://github.com/Pocket/curation-admin-tools/assets/22447785/f4a75757-8cd9-455a-b3a6-ab87d35eef54

(shows the same story repeatedly scheduled on one surface, once with a 2-day gap, which is highlighted, and then switches to another surface where the same 2-day gap is not highlighted because it relates to a different surface)

## Reference

https://mozilla-hub.atlassian.net/browse/MC-572


package.json
```diff
@@ -40,7 +40,7 @@
     "yup": "0.32.11"
   },
   "scripts": {
-    "start": "NODE_ENV=local react-scripts start",
+    "start": "NODE_OPTIONS=--openssl-legacy-provider NODE_ENV=local react-scripts start",
```
nina-py, line 43: This fixes the "digital envelope routines" error for me

src/_shared/components/CorpusItemCardImage/CorpusItemCardImage.tsx
```diff
@@ -61,7 +66,7 @@ const bottomOverlayContainerSxStyles: SxProps = {
   cursor: 'pointer',
 };
 
-const getTopRightLabel = (item: ApprovedCorpusItem): JSX.Element | null => {
+const getTopRightLabel = (item: ApprovedCorpusItem): ReactElement | null => {
```
nina-py, line 69: Getting rid of deprecation notices as I touch these files

jpetto (approved):

Herraj (approved):
🚀

katerinachinnappan (approved):

PR #1167, opened 2024-02-21
refactor: uppercase "Other" removal reason for consistency

jpetto:
## Goal

make the "Other" checkbox send to the API in all caps for consistency with the enum values.


katerinachinnappan (approved):
🔥🥳

PR #1170, opened 2024-02-28
feat: send 'MANUAL' as scheduled item source

jpetto:
## Goal

send `MANUAL` as the `source` when creating scheduled items.

- fix up missing isMlScheduled prop issues found along the way

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-752

## Implementation Decisions

while the `createApprovedCorpusItem` mutation can [accept scheduling info](https://github.com/Pocket/content-monorepo/blob/main/servers/curated-corpus-api/schema-admin.graphql#L638-L645) (`scheduledDate` and `scheduledSurfaceGuid`), i don't see any usage of scheduling through this mutation in the admin tools. am i missing something? why does that mutation accept scheduling info? (i should know this but have forgotten.)


nina-py (approved):
🔥

PR #1174, opened 2024-03-05
fix: adding source to mutation when moving item to bottom

katerinachinnappan:
## Goal
adding `source` to reschedule mutation when moving item to bottom.

### DIFF
--- a/src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
+++ b/src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
@@ -333,9 +333,11 @@ export const SchedulePage: React.FC = (): ReactElement => {
   const moveItemToBottom = (item: ScheduledCorpusItem): void => {
     // Run the "reschedule" mutation with the same variables
     // it already has. All we want from it is to update `updatedAt` timestamp.
+    console.log('move to bottom');
     const variables = {
       externalId: item.externalId,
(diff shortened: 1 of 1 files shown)


src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -333,9 +333,11 @@ export const SchedulePage: React.FC = (): ReactElement => {
   const moveItemToBottom = (item: ScheduledCorpusItem): void => {
     // Run the "reschedule" mutation with the same variables
     // it already has. All we want from it is to update `updatedAt` timestamp.
+    console.log('move to bottom');
```
jpetto, line 336: can we remove this 😅

src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -333,9 +333,11 @@ export const SchedulePage: React.FC = (): ReactElement => {
...
     const variables = {
       externalId: item.externalId,
       scheduledDate: item.scheduledDate,
+      source: ScheduledItemSource.Ml,
```
jpetto, line 340: this should be `Manual`, right? what a monday!

jpetto (changes requested):

jpetto (approved):

PR #1175, opened 2024-03-05
fix(scheduling): Add source field to query fragment; update generated types

nina-py:
## Goal

- This removes a TypeScript error from the SchedulePage component.

Follow-up fix to https://mozilla-hub.atlassian.net/browse/MC-763.

### DIFF
--- a/src/api/queries/getScheduledItems.ts
+++ b/src/api/queries/getScheduledItems.ts
@@ -1,5 +1,5 @@
 import { gql } from '@apollo/client';
-import { CuratedItemData } from '../fragments/curatedItemData';
+import { ScheduledItemData } from '../fragments/scheduledItemData';
 
 export const getScheduledItems = gql`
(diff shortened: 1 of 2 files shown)


src/api/queries/getScheduledItems.ts
```diff
@@ -9,18 +9,9 @@ export const getScheduledItems = gql`
...
-        approvedItem {
-          ...CuratedItemData
-        }
+        ...ScheduledItemData
       }
     }
   }
-  ${CuratedItemData}
+  ${ScheduledItemData}
 `;
```
nina-py, line 17: Updated to use the `ScheduledItemData` which had the `source` field added in the parent ticket. I _think_, initially, this query didn't pull in all the fields from scheduled items, which is why fields were listed individually instead of using the fragment with all the available fields.

katerinachinnappan (approved):

jpetto (approved):
ty nina!

PR #1176, opened 2024-03-05
feat(scheduling): [MC-784] 'Add item' button should appear next to each date on Schedule page

nina-py:
## Goal

As currently implemented, the “Add item” button on the Schedule page is a solitary button on top of the page, and the scheduling date it fills in by default is tomorrow in the curator’s time zone. This was actually not the intention of the new UX wireframes - there should be a button next to each scheduled date. 

Functionality changes:

- There is an “Add an item” for manual addition of stories alongside each date displayed on the Schedule page

- When going through the manual addition, the date automatically filled in should be that parent date rather than tomorrow.

## Implementation details

- I have slightly de-emphasised the header for each date and lined up the header/accordion element with the button. Note that the accordion is on the way out - it will be replaced by the new sorting and filtering in the near future.

## Video walkthrough

Shows the multiple "Add item" buttons and the date the screen defaults to when scheduling items (in this case, March 21).

https://github.com/Pocket/curation-admin-tools/assets/22447785/732b87d7-651f-4bd2-98a0-f29728314ebd

## Reference

https://mozilla-hub.atlassian.net/browse/MC-784

### DIFF
--- a/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
+++ b/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
@@ -70,6 +70,7 @@ export const ScheduleItemForm: React.FC<
     onSubmit,
   } = props;
 
+  // Default to tomorrow if no selected date is passed through the props.
   const tomorrow = DateTime.local().plus({ days: 1 });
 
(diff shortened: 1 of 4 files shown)


jpetto (approved):
this is nice!

PR #1180, opened 2024-03-26
feat(scheduling): [MC-781] Add publication date to cards in Schedule View

nina-py:
## Goal

- Added the `datePublished` field throughout to be able to add this data when manually adding prospects or curated items. This is a hidden field and the data is supplied by the Pocket Graph for collections and syndicated items and by the Parser for all other stories if available.

- Added the publication date to the new look scheduled item card.

- Updated tests, query and mutation mocks, generated types.

## Implementation decisions

- The publication date is a hidden field for the time being - if the curators need to adjust or view it when adding a curated item manually, we can find a spot for it on the Approved Item form. I didn't want to add any more scope creep to a ticket that started out as a seemingly very small task.

## Video walkthrough

I'm adding an item manually on the Schedule page in the walkthrough; adding an item manually from the Prospecting page also works similarly - in that, it adds a date if it's available.

https://github.com/Pocket/curation-admin-tools/assets/22447785/cdd175a7-a4f3-4dd3-8914-d87f69660783

## Reference

Ref: https://mozilla-hub.atlassian.net/browse/MC-781

### DIFF
--- a/src/api/fragments/curatedItemData.ts
+++ b/src/api/fragments/curatedItemData.ts
@@ -10,6 +10,7 @@ export const CuratedItemData = gql`
     title
     language
     publisher
+    datePublished
     authors {
       name
       sortOrder
--- a/src/api/fragments/urlMetadata.ts
+++ b/src/api/fragments/urlMetadata.ts
@@ -8,6 +8,7 @@ export const urlMetadata = gql`
     url
     imageUrl
     publisher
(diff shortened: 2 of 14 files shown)


jpetto (approved):
looking good! ty!

katerinachinnappan (approved):
🔥

PR #1182, opened 2024-03-27
fix(corpus): Adjust validation for 'datePublished' field to allow for null values

nina-py:
## Goal

This is a follow-up fix to https://github.com/Pocket/curation-admin-tools/pull/1180 which broke editing metadata for existing curated items that don't have a publication date.


mmiermans (approved):
Thanks for the quick fix!

PR #1183, opened 2024-04-04
feat(scheduling): [MC-523] Curators must provide a reason for manuall…  …y scheduling items

nina-py:
## Goal

- Added reasons for manually scheduling items + comment field for "Other" reasons to scheduling form. These only appear if requested, and are required if they appear on the page.

- Reorganised the elements on the Scheduling modal, so that surface and date are now on a single line, and topic & publisher distribution is rolled up by default in an accordion element.

- Made sure manually added items from the Prospecting page also have a list of reasons to choose from, identical to the Scheduling page.

- Added tests.

## Screenshots

The reworked schedule modal:

reworked-schedule-modal

For other surfaces, it looks more compact without the reasons:

schedule-modal-non-us

Expanded topic & publisher distribution:

schedule-modal-non-us-expanded

## Video walkthrough

https://github.com/Pocket/curation-admin-tools/assets/22447785/26300366-db36-403a-a1cb-23bec4fe49a7

## Reference

https://mozilla-hub.atlassian.net/browse/MC-523

PR #1184, opened 2024-04-05
fix(scheduling): [MC-922] Do not require manual scheduling reasons for syndicated items

nina-py:
## Goal

Modify the workflow introduced in https://github.com/Pocket/curation-admin-tools/pull/1183 so that our syndication coordinator does not need to provide reasons when manually adding syndicated articles to the corpus.

## Video walkthrough

https://github.com/Pocket/curation-admin-tools/assets/22447785/ce295b8a-eb4c-4b04-859e-22f924880e20

## Reference

https://mozilla-hub.atlassian.net/browse/MC-922

### DIFF
--- a/src/curated-corpus/components/actions/ScheduleCorpusItemAction/ScheduleCorpusItemAction.test.tsx
+++ b/src/curated-corpus/components/actions/ScheduleCorpusItemAction/ScheduleCorpusItemAction.test.tsx
@@ -48,13 +48,15 @@ describe('The ScheduleCorpusItemAction', () => {
   it('completes the action successfully', async () => {
     // setting a fixed timezone since this test has a tendency of passing locally but failing
     // on CI due to server and local timezone discrepancies
-    const today = DateTime.local()
-      .setZone('America/New_York')
+    const chosenDate = DateTime.local()
+      .setZone('Europe/London')
       .plus({ days: 1 });
+
+    const chosenSurfaceGuid = 'NEW_TAB_EN_GB';
 
     mocks = [
       mock_AllScheduledSurfaces,
-      successMock(today.toFormat('yyyy-MM-dd')),
+      successMock(chosenDate.toFormat('yyyy-MM-dd'), chosenSurfaceGuid),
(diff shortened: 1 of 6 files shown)


jpetto (dismissed):

src/curated-corpus/integration-test-mocks/getScheduledSurfacesForUser.ts
```diff
@@ -25,7 +24,6 @@ const allScheduledSurfaces: ScheduledSurface[] = [
       ProspectType.TitleUrlModeled,
       ProspectType.RssLogistic,
       ProspectType.RssLogisticRecent,
-      ProspectType.SlateScheduler,
       ProspectType.SlateSchedulerV2,
```
nina-py, line 27: Unrelated to the main purpose of this PR - these two prospect types are no longer valid, removed them from the mocks

src/curated-corpus/components/actions/ScheduleCorpusItemAction/ScheduleCorpusItemAction.test.tsx
```diff
@@ -48,13 +48,15 @@ describe('The ScheduleCorpusItemAction', () => {
...
     // on CI due to server and local timezone discrepancies
-    const today = DateTime.local()
-      .setZone('America/New_York')
+    const chosenDate = DateTime.local()
+      .setZone('Europe/London')
       .plus({ days: 1 });
 
+    const chosenSurfaceGuid = 'NEW_TAB_EN_GB';
+
     mocks = [
```
nina-py, line 57: Updated this test with more meaningful variable names (since now() + one day was never "today" - this was a remnant of the very first iteration of this test), and a little bit more customisation of the mocks.

Also tried experimenting with setting a time zone in in CircleCI jobs, but it didn't make any difference, so I reverted those changes.

jpetto (approved):
this is looking good - and ty for the clean up. i hope that timezone in circle does the trick 😅

PR #1185, opened 2024-04-05
feat: send actionScreen value to relevant mutations

jpetto:
## Goal

send actionScreen value to relevant mutations.

## Reference

- https://mozilla-hub.atlassian.net/browse/MC-748


nina-py (approved):
🚀

PR #1190, opened 2024-04-15
feat(scheduling): [MC-923] Expand Topic & Publisher summary for manually scheduled syndicated articles

nina-py:
## Goal

Now that the Topic & Publisher summary is hidden by default in an Accordion component to make space for the manual  schedule reasons, let's expand it for syndicated articles to simplify curators' workflow. 

## Video Walkthrough

https://github.com/Pocket/curation-admin-tools/assets/22447785/99e378d9-bf4e-4b0c-a462-70bd4da91d2e

## Reference

https://mozilla-hub.atlassian.net/browse/MC-923

### DIFF
--- a/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
+++ b/src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
@@ -68,6 +68,11 @@ interface ScheduleItemFormProps {
   showManualScheduleReasons?: boolean;
 
   /**
+   * Whether to expand the Topic & Publisher summary.
+   */
+  expandSummary?: boolean;
+
+  /**
    *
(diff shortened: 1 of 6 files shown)


src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
```diff
@@ -227,7 +233,7 @@ export const ScheduleItemForm: React.FC<
           >
             <Grid item xs={12}>
               <Box mt={3}>
-                <Accordion>
+                <Accordion defaultExpanded={expandSummary}>
```
nina-py, line 236: This setting lets us expand the Accordion component if necessary without having to manage its state manually.

mmiermans (approved):
We :pear:'ed on the last commit.

PR #1191, opened 2024-04-16
fix(scheduling): [MC-962] Reschedule button moves the last article posted

nina-py:
## Goal

Fix the bug - rescheduling modal should be showing the correct item and not the latest (manual) item added to the schedule. 

(invalid - was thinking of reordering items rather than rescheduling them) ~~Fix the bug! I have traced this bug to April 2022: https://github.com/Pocket/curation-admin-tools/commit/92d14a6e15e8ef2b9c86c98f9ca126f3c038098e#r141000157 - which means it has been there ever since this feature was introduced, it may not have been used often enough to be detected up until recently.~~

## Video walkthrough

Adding a manual item, then rescheduling a different item - all works as intended now.

https://github.com/Pocket/curation-admin-tools/assets/22447785/bc87346f-737e-488c-98ba-eb641e8f4c98

## Reference

https://mozilla-hub.atlassian.net/browse/MC-962

### DIFF
--- a/src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
+++ b/src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
@@ -125,6 +125,13 @@ export const SchedulePage: React.FC = (): ReactElement => {
   const [scheduleItemModalOpen, toggleScheduleItemModal] = useToggle(false);
 
   /**
+   * Set up a separate toggle for an identical modal that is used for
+   * manually scheduled items.
+   */
+  const [manualScheduleItemModalOpen, toggleManualScheduleItemModal] =
+    useToggle(false);
+
+  /**
    * Keep track of whether the "Edit item modal" is open or not
    */
   const [editItemModalOpen, toggleEditModal] = useToggle(false);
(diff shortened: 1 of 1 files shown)


src/curated-corpus/pages/SchedulePage/SchedulePage.tsx
```diff
@@ -557,13 +564,7 @@ export const SchedulePage: React.FC = (): ReactElement => {
...
 
         // Hide the Schedule Item Form modal
-        toggleScheduleItemModal();
-
-        // Unset the corpus item that was added manually so that
-        // the ScheduleItemModal specific to manually added stories
-        // (with reasons for scheduling for the New Tab (US) surface)
-        // does not show up when a reorder/reschedule action is triggered.
-        setApprovedItem(undefined);
+        toggleManualScheduleItemModal();
```
nina-py, line 567: Unsetting the last worked on approved item is no longer needed as we're not using the same on/off toggle for the Schedule Item modal any more, so this can be taken out. It was a half-way fix, but it would not have worked consistently all the time.

jpetto (approved):
so much to keep track of on this page now 😅. ty for the fix!

PR #1193, opened 2024-04-23
MC-858: Provide manual reasons for manually adding items on German New Tab

katerinachinnappan:
## Goal

- When adding an item Manually  from Prospecting for the New Tab DE-de surface, the curators will be prompted to add a Reason.

- When adding an item Manually  from the Schedule page for the New Tab DE-de surface, the curators will be prompted to add a Reason.

## Todos

- [x] deployed to dev

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-858


jpetto (approved):
👏

PR #1194, opened 2024-04-27
feat: display new domain warning

mmiermans:
## Goal
Display a warning on the schedule page for items that have not been scheduled in the past.

## Deployment

- [x] Deploy to production: https://github.com/Pocket/content-monorepo/pull/155

## Reference

Design:

- https://mozilla-hub.atlassian.net/browse/MC-832?focusedCommentId=858103

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-908

Documentation:

- [Solutions that were considered](https://mozilla-hub.atlassian.net/wiki/spaces/MozSocial/pages/656277652/New+domain+warning+on+schedule)
- [Rapid Risk Assessment (RRA) for New Tab Scheduler](https://docs.google.com/document/d/1T79X5rOcUHY_Go7bH9ijlP5kg35Ou2VSqnIC01I1JeQ/edit#heading=h.ufs6i6416jwo)
- [Eric's RRA Feedback wrap-up](https://docs.google.com/document/d/1UhPfwsdWuB_CK-1EfzM42ZKiEt3zJiw5N-VwjCTeA5g/edit#heading=h.wy1yuau5f6bf)

## QA

image

PR #1196, opened 2024-04-30
fix(tests): [MC-1021] Fix flaky test for scheduled item action

nina-py:
## Goal

The 'happy path' test for `ScheduleCorpusItemAction` has been flaky for ages. Multiple attempts have been made to fix this. We paired with @mmiermans yesterday on https://github.com/Pocket/curation-admin-tools/pull/1194 and seem to have found a way to fix it within the Jest environment.

I combed over the code and discovered that two components were setting a default date where only one really should, and once I took out the redundant code the query for Topic & Publisher distribution that is part of the `ScheduleItemForm`, the last of the four components in the chain (`ScheduleCorpusItemAction` -> `ScheduleItemModal` -> `ScheduleItemFormConnector` -> `ScheduleItemForm`), started running in the test. I added a mock for that.

I've also manually verified that scheduling works throughout the app.

## Reference

https://mozilla-hub.atlassian.net/browse/MC-1021


src/curated-corpus/components/ScheduleItemForm/ScheduleItemForm.tsx
```diff
@@ -113,7 +112,7 @@ export const ScheduleItemForm: React.FC<
 
   useEffect(() => {
     setRefreshData(true);
-    handleDateChange(selectedDate ? selectedDate : tomorrow);
+    handleDateChange(selectedDate);
```
nina-py, line 115: Don't set a default date in the form as we already do so in the component above it (ScheduleItemFormConnector), and the form component is never used anywhere directly.

src/curated-corpus/components/actions/ScheduleCorpusItemAction/ScheduleCorpusItemAction.test.tsx
```diff
@@ -56,6 +57,10 @@ describe('The ScheduleCorpusItemAction', () => {
 
     mocks = [
       mock_AllScheduledSurfaces,
+      mock_scheduledItemsWithParams(
+        chosenDate.toFormat('yyyy-MM-dd'),
+        chosenSurfaceGuid
+      ),
```
nina-py, line 63: This is a mock for the Topic & Publisher info that was unreachable before in this form (???), but is needed now.

jpetto (approved):
nice!

PR #1198, opened 2024-05-06
chore(dependencies): [MC-1163] Eject react scripts

nina-py:
## Goal

Start the process of updating this app's many outdated or deprecated dependencies, some of which are responsible for a record 43 security alerts on this repository.

Todos:

- [x] Run `npm eject`, which uninstalled `react-scripts` package (Create React App, copied over all the configuration files into the app's directory and exposed all the package dependencies, e.g. Webpack, directly in the `package.json` file. 
- [x] Updated ESLint to v8 and some of its plugins, too
- [x] Updated the two GitHub actions (linting and unit/integration tests) to run on Node 18. Unfortunately updating _all_ of the infrastructure to run on Node 18, never mind Node 20, is not yet possible - the current version of Webpack (v4) only supports node up to v16.
- [x] Fixed linting errors on `*.ts` files that somehow were out of scope for the linter before, and also some in the `.aws` directory. 
- [x] Deploy to Dev
- [x] Test on Dev

## Notes

The "Run Unit Tests (16.x)" check has been superseded by the "Curation Admin Tools / Run Unit Tests (18.x) (pull_request)" check - this is in fact the same job, upgraded to run in Node 18 environment. Once this PR is merged, repository settings need to be updated to make the new (18x) check required. 

## Reference

https://mozilla-hub.atlassian.net/browse/MC-1163


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # development build - build app
-FROM node:16@sha256:466d0a05ecb1e5b9890960592311fa10c2bc6012fc27dbfdcc74abf10fc324fc as builder
+FROM node:16 as builder
```
nina-py, line 2: This is the part where we cannot upgrade even to Node 18 yet as Webpack v4 holds us back at Node 16 for the time being.

jpetto: 😭

config/env.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
```
jpetto, line 1: i'm guessing these scripts in `config/` are the result of the ejection and not custom written? do we know that these scripts are necessary to keep/maintain going forward, or is their presence more of a transitional step?

package.json
```diff
@@ -15,41 +16,142 @@
...
+    "eslint-plugin-react-hooks": "4.6.2",
+    "eslint-plugin-testing-library": "6.2.2",
+    "eslint-webpack-plugin": "2.7.0",
+    "file-loader": "6.1.1",
     "formik": "2.2.9",
+    "fs-extra": "^9.0.1",
+    "html-webpack-plugin": "4.5.0",
+    "identity-obj-proxy": "3.0.0",
+    "jest": "26.6.0",
+    "jest-circus": "26.6.0",
```
jpetto, line 54: this is a circus! 🎪

jpetto (approved):
well, i didn't review all the (i assume) ejected scripts for correctness as that would take forever 🙃. i think testing this on dev is sufficient.

biggest question here is - will we need to retain and maintain all these ejected scripts? are we...prepared for that? 😅 (not arguing against ejecting/moving off create react app, just want to know what our future holds.)

config/env.js
nina-py: Yes, the config files are added to the app directory once the app is ejected. Most of these are for Webpack and Jest. And yes, we'll have to maintain them going forward, which is not as arduous as I thought it would be - the files committed in this PR have already been tweaked slightly to make sure all functionality works.

PR #1201, opened 2024-05-22
feat(scheduling): [MC-669] Allow filtering in Schedule View

nina-py:
## Goal

Add filters to each day's data on the Schedule page. Wireframes in ticket.

Acceptance Criteria (copied over from JIRA):

- For a given day on the schedule, allow Filtering by Topic, Publisher, or Type 
- Provide Filtering for each day on the page separately (for example, if I am looking at the 24th & 25th on the page at once I can use the filter for the 24th or filter the 25th)
- Changing the date resets all of the filters
- Only one filter is available at a time

## Video Walkthrough

Note: PR currently deployed to Dev, so can be test-driven there: https://curation-admin-tools.getpocket.dev/

https://github.com/Pocket/curation-admin-tools/assets/22447785/310d3799-99bd-4f6d-afc5-86e71a2a8d50

## References

JIRA: https://mozilla-hub.atlassian.net/browse/MC-669


src/curated-corpus/components/ScheduleDayData/ScheduleDayData.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Show all scheduled entries for a given date.
+ *
+ * @param props
+ * @constructor
+ */
+export const ScheduleDayData: React.FC<ScheduleDayDataProps> = (
+  props,
+): ReactElement => {
```
nina-py, line 76: This component is a big chunk of functionality from SchedulePage, now in its own component. It's still too cumbersome to test effectively - I think it needs to be broken up further for that. The other three new components have test coverage.

src/_shared/components/CardActionButtonRow/CardActionButtonRow.tsx
```diff
@@ -36,7 +36,7 @@ interface CardActionButtonRowProps {
 }
 
 export const CardActionButtonRow: React.FC<CardActionButtonRowProps> = (
-  props
+  props,
```
nina-py, line 39: A whole heap of linting tweaks were needed for the app to run locally - strange, but ok.

jpetto: i just ran into this too! weird...

jpetto (approved):
this looks good to me. i tested on dev too and everything is working as expected.

i kind of expected the filters to show the selection, but if curators are good with it as-is, i don't think we need to wait. (maybe showing the selection can be a follow-up.)

PR #1203, opened 2024-06-04
MC-1123: Support for "Fix Title", quotes for German Language via UI

katerinachinnappan:
## Goal

* There should be no change in capitalization for title

* Quotation marks need to be updated to „ at the beginning and “ at the end.

* Dashes should be long (em dash) ( – ) not short ( - )

## Deployment steps

- [ ] Deployed to dev

## References

JIRA ticket: https://mozilla-hub.atlassian.net/browse/MC-1123

### DIFF
--- a/src/_shared/utils/applyQuotesDashesDE.ts
+++ b/src/_shared/utils/applyQuotesDashesDE.ts
@@ -1,0 +1,21 @@
+/**
+ * Helper to replace opening and closing quotes, arrow brackets (» «) for German
+ * Quotes represented in unicode
+ *
+ * @param text
+ * @returns string
+ */
+
+export const applyQuotesDashesDE = (text: string): string | undefined => {
+  if (!text) {
+    return undefined;
+  }
+  return text
+    .replace(/(^|[-\u2014/([{\u2018\s])\u00AB/g, '$1\u201E') // Replaces opening « with „
+    .replace(/\u00BB/g, '\u201D') // Replaces closing » with ”
+    .replace(/(^|[-\u2014/([{\u2018\s])"/g, '$1\u201E') // Opening doubles (replaces opening " with „)
+    .replace(/"/g, '\u201D') // Closing doubles (replaces closing " with ”)
+    .replace(/(^|[-\u2014/([{\u2018\s])\u201c/g, '$1\u201E') // Replaces opening “ with „
+    .replace(/\s\u2013\s/g, ' \u2014 ') // Replace en dash (–) with long em dash (—)
+    .replace(/\s-\s/g, ' \u2014 '); // Replace short dash (-) with long em dash (—)
+};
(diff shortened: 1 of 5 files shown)


jpetto (changes requested):
one request to create separate functions for formatting titles and excerpts, but otherwise looks good!

src/_shared/utils/applyQuotesDashesDE.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Helper to replace opening and closing quotes, arrow brackets (» «) for German
+ * Quotes represented in unicode
+ *
+ * @param text
+ * @returns string
+ */
+
+export const applyQuotesDashesDE = (text: string): string | undefined => {
```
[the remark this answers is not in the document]
katerinachinnappan: yep, once we have it merged into the monorepo 😄

src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -175,3 +181,51 @@ export const getLastScheduledDayDiff = (
...
+ * @param excerpt excerpt to format
+ * @returns formatted excerpt
+ */
+export const applyExcerptFormattingByLanguage = (
+  language: CorpusLanguage,
+  excerpt: string,
+): string => {
+  // if excerpt is English, apply quotes formatting for EN
+  if (language === CorpusLanguage.En) {
+    applyCurlyQuotes(excerpt);
```
jpetto, line 197: missing a `return` here! (but see below for more refactoring.)

also - this function still works as expected for EN, even with this missing `return`. but why? 😉

src/curated-corpus/helpers/helperFunctions.ts
```diff
@@ -175,3 +181,51 @@ export const getLastScheduledDayDiff = (
...
+  }
+  // if excerpt is German, apply German quotes/dashes formatting
+  if (language === CorpusLanguage.De) {
+    return applyQuotesDashesDE(excerpt) as string;
+  }
+  // apply EN formatting rules on excerpt for other languages for now
+  else {
+    return applyCurlyQuotes(excerpt);
+  }
```
jpetto, line 206: just a little cleanu. this function could be simplified to:

```
// DE is the only special case for now
if (langauge === CorpusLanguage.De) {
  return applyQuotesDashesDE(excerpt) as string;
// all other languages get the EN treatment
} else {
  return applyCurlyQuotes(excerpt);
}
```

same pattern for the function below.

jpetto (changes requested):
last little bit of cleanup 😅

src/curated-corpus/helpers/helperFunctions.ts
katerinachinnappan: oo good catch! its always switching to the `else` block as nothing is explicitly returned in the EN `if` block. & they both return the same result

jpetto (approved):
looking good!

PR #1205, opened 2024-06-13
MC-1190: fix German quotes / dashes formatting

katerinachinnappan:
## Goal

- Dashes should be changed to `en` dashes not `em` dashes. Fixed & updated tests.
- Accommodate »example« use case, should be changed to „example"

## Deployment steps

- [x] Deployed to dev

## References

JIRA ticket:

- https://mozilla-hub.atlassian.net/browse/MC-1190


jpetto (approved):
🇩🇪

PR #1207, opened 2024-06-18
fix(linting): fix local npm lint scripts

jpetto:
## Goal

fix local npm lint scripts

- also run lint-fix and add the fixed files

## Test Locally

pull down the latest `main` and run `npm run lint-check`. for me, this reported one error in a file in `node_modules` - which is a directory it shouldn't even check. it also does not report the errors that my fix here does.

still on the `main` branch, update the `package.json` lint scripts to match the ones in this PR and run them. you should get the correct output (like 80 files needing fixing).

## Implementation Decisions

does anyone know why we were running `tsc` before the lint scripts? why would we need to compile to only lint `.ts(x)` files?

also - unclear why the files caught _here_ are no being caught by the github linter. other repositories are experiencing similar local/github linting discrepancies - wonder if some downstream package got changed...

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,8 +97,8 @@
     "build": "DISABLE_ESLINT_PLUGIN=true node scripts/build.js",
     "test": "node scripts/test.js --transformIgnorePatterns \"node_modules/(?!@apollo/client/link/core/ApolloLink.js)/\"",
     "test:nowatch": "node scripts/test.js --watchAll=false  --transformIgnorePatterns \"node_modules/(?!@apollo/client/link/core/ApolloLink.js)/\"",
-    "lint-check": "tsc --noEmit && eslint src/**/*.ts{,x}",
-    "lint-fix": "tsc --noEmit && eslint src/**/*.ts{,x} --fix",
+    "lint-check": "eslint src/**/*.ts{,x}",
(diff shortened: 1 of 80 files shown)


package.json
```diff
@@ -97,8 +97,8 @@
     "build": "DISABLE_ESLINT_PLUGIN=true node scripts/build.js",
     "test": "node scripts/test.js --transformIgnorePatterns \"node_modules/(?!@apollo/client/link/core/ApolloLink.js)/\"",
     "test:nowatch": "node scripts/test.js --watchAll=false  --transformIgnorePatterns \"node_modules/(?!@apollo/client/link/core/ApolloLink.js)/\"",
-    "lint-check": "tsc --noEmit && eslint src/**/*.ts{,x}",
-    "lint-fix": "tsc --noEmit && eslint src/**/*.ts{,x} --fix",
+    "lint-check": "eslint src/**/*.ts{,x}",
+    "lint-fix": "eslint src/**/*.ts{,x} --fix",
```
nina-py, line 101: This would have been set up by us manually back in the day AFAIK, perhaps with the ejection from `react-scripts` and major version update to ESLint this is no longer needed?

nina-py (approved):

katerinachinnappan (approved):

PR #1209, opened 2024-07-09
feat: add grade buttons to scheduled item cards

jpetto:
## Goal

add grade buttons to scheduled item cards.

## Todos

- [ ] push the [backend PR](https://github.com/Pocket/content-monorepo/pull/185) to dev and test the full flow of data to snowplow
- [ ] add tests for the new buttons?

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-1156

### DIFF
--- a/src/curated-corpus/components/ScheduleDayData/ScheduleDayData.tsx
+++ b/src/curated-corpus/components/ScheduleDayData/ScheduleDayData.tsx
@@ -1,10 +1,14 @@
 import React, { ReactElement, useState } from 'react';
 import { Box, Grid } from '@mui/material';
 import {
+  ActionScreen,
+  ApprovedCorpusItem,
+  ApprovedItemGrade,
   ScheduledCorpusItem,
   ScheduledCorpusItemsResult,
   ScheduledItemSource,
   useRescheduleScheduledCorpusItemMutation,
+  useUpdateApprovedCorpusItemGradeMutation,
 } from '../../../api/generatedTypes';
 import {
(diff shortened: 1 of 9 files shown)


src/curated-corpus/components/ScheduleDayData/ScheduleDayData.tsx
```diff
@@ -126,6 +132,36 @@ export const ScheduleDayData: React.FC<ScheduleDayDataProps> = (
...
+
+    runMutation(
+      updateCorpusItemGrade,
+      { variables },
+      `Curated item "${item.title.substring(0, 40)}..." successfully graded`,
+      undefined,
+      undefined,
+      refetch,
+    );
+  };
```
jpetto, line 163: is this a good place to define this function?

the next PR for this will be to add the grade field to the item edit modal on all pages - which makes me think this function should be abstracted at a higher level, but i don't see an existing pattern for such a thing.

katerinachinnappan: I think if you look at the existing pattern & function above this, `moveItemToBottom`, you can leave it here

nina-py: I think it's fine. This component is where a big chunk of SchedulePage was moved out to. The plan is to split it further into testable components some time in the future 🤔

jpetto:
closing as we are not implementing grades

PR #1210, opened 2024-07-12
chore(packages): [MC-1308] Upgrade notistack library to v2

nina-py:
## Goal

Upgrade this library so that we no longer have an indirect dependency on Material UI v4 which has not been in active development since 2021.

- Upgraded the `notistack` library to v.2.0.8
- Removed one last remaining dependency on Material UI 4 that was left over in our own code
- Tests are passing; additionally, manually tested that the duplicate item modal and notices work fine

## Reference

https://mozilla-hub.atlassian.net/browse/MC-1308


src/curated-corpus/components/DuplicateProspectModal/DuplicateProspectModal.styles.tsx
```diff
@@ -1,25 +0,0 @@
...
- */
-export const useStyles = makeStyles((theme: Theme) =>
-  createStyles({
-    root: {
-      width: '800px',
-    },
-    link: {
-      color: theme.palette.primary.main,
-      textDecoration: 'none',
-    },
```
nina-py, line 14: This style wasn't needed in the duplicate item modal - I think an earlier version of it had a link to the corpus item page, but it has since been replaced with a button. The rest of the styles have been moved over to the component itself into `sx` props.

katerinachinnappan (approved):
Ty!

PR #1212, opened 2024-08-02
feat(topics): Add new 'Home' topic as an option in the tool

nina-py:
# DO NOT DEPLOY

See notes in https://mozilla-hub.atlassian.net/browse/MC-1248 for when this feature is going to production.

## Goal

Allow the curators to categorise stories as belonging to the brand new "Home" category. 

## Implementation details

- Note: with regenerating types, a handful of unrelated schema changes have sneaked through. I think that is fine - we'll use them in the near future anyway. It's also possible that these schema changes will be merged to `main` earlier than this PR is deployed, in which case the changes will be removed from this PR.

### Tested manually:

- [x] Adding a new corpus item manually with "Home" as topic
- [x] Editing an existing item to set the topic to "Home"
- [x] Adding a prospect to the corpus with "Home" as topic
- [x] Scheduled item cards - topic is displayed correctly (see screenshot below)

home-article

## Reference

https://mozilla-hub.atlassian.net/browse/MC-1251

### DIFF
--- a/src/curated-corpus/helpers/topics.test.ts
+++ b/src/curated-corpus/helpers/topics.test.ts
@@ -43,7 +43,7 @@ describe('helper functions related to topics', () => {
       expect(topics).to.be.an('array');
       // However, we're getting the entire list of topics here,
       // not just the ones stories belong to
-      expect(topics).to.have.lengthOf(16);
+      expect(topics).to.have.lengthOf(17);
 
       // And we expect to see the story topics, with the correct counts
       expect(topics).to.contain.deep.members([
--- a/src/curated-corpus/helpers/topics.test.ts
+++ b/src/curated-corpus/helpers/topics.test.ts
@@ -89,7 +89,7 @@ describe('helper functions related to topics', () => {
 
     it('returns a list of topics without "Coronavirus" if requested', () => {
       const topics = getGroupedTopicData(data, true, false);
-      expect(topics).to.be.an('array').with.lengthOf(15);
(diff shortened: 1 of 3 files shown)


src/curated-corpus/helpers/topics.test.ts
```diff
@@ -89,12 +89,12 @@ describe('helper functions related to topics', () => {
...
     it('returns a list of topics without "Coronavirus" if requested', () => {
       const topics = getGroupedTopicData(data, true, false);
-      expect(topics).to.be.an('array').with.lengthOf(15);
+      expect(topics).to.be.an('array').with.lengthOf(16);
     });
 
     it('returns a full list of topics if requested', () => {
       const topics = getGroupedTopicData(data);
-      expect(topics).to.be.an('array').with.lengthOf(16);
+      expect(topics).to.be.an('array').with.lengthOf(17);
```
nina-py, line 97: With most of the test updates involving updating the number of topics expected, should tests be modified to not check for the total number of topics? Should we compare them to the enums in the generated code instead? 

Or is adding a topic going to be a rare enough event that it's fine to leave this - and by extension the process of adding a topic - as-is?

jpetto: good question and i'm not sure we know the answer yet re: how often will we get new topics. but - i think your point about avoiding testing a specific number of topics is sound. we should stop testing the number - i think testing against the enum sounds good.

jpetto (approved):

PR #1213, opened 2024-08-05
fix(filters): [MC-1273] Make sure ML-filtered items don't include Syndicated items

nina-py:
## Goal

It appears the initial PR that added filters to the Schedule page did not filter out ML-Syndicated items out of the ML filter 
correctly. The dropdown numbers shown were correct, however the actual filters were not applied correctly. 

I _think_ this is because we didn't have enough data to test on Dev with, and that is why the bug must have slipped through. With the data we had at the time, everything looked to be working as expected. 
## Reference

https://mozilla-hub.atlassian.net/browse/MC-1273?focusedCommentId=919426


jpetto (approved):
nice catch!

PR #1215, opened 2024-09-03
Mc 1404 create prospecting source for publisher air table suggestions

rolf-moz:
## Goal

What changed? What is the business/product goal?
The Editors would like to have easy access to stories suggested by external publishers. 
They are currently being submitted via airtable.

We already have ingestion in our German auto-scheduler, so we are leveraging that to output a new prospect feed in this PR
https://github.com/MozillaSocial/ml-services/pull/430

New type is added in https://github.com/Pocket/content-monorepo/pull/213

References to deprecated Syndication prospect feed are removed as well.

## Implementation Decisions

Named the prospect source ExternalEditorial so users know these are external links and thus unvetted. 

Only URLs from previously published domains are sent to this feed. We might consider additional security checks in the future.

## Deployment steps
- [x] Deployed to dev?

## References

JIRA ticket:
[
- Link to JIRA ticket](https://mozilla-hub.atlassian.net/browse/MC-1404)

Issue:

- Link to GitHub issue

Documentation:

- Project doc
- https://mozilla-hub.atlassian.net/wiki/spaces/MozSocial/pages/762642732/Airtable+Editor+Submission+Specification

### DIFF
--- a/src/api/generatedTypes.ts
+++ b/src/api/generatedTypes.ts
@@ -1562,7 +1562,7 @@ export enum ProspectType {
   CountsModeled = 'COUNTS_MODELED',
   Dismissed = 'DISMISSED',
   DomainAllowlist = 'DOMAIN_ALLOWLIST',
-  ExternalEditorial = 'EXTERNAL_EDITORIAL',
+  PublisherSubmitted = 'PUBLISHER_SUBMITTED',
   Recommended = 'RECOMMENDED',
   RssLogistic = 'RSS_LOGISTIC',
   RssLogisticRecent = 'RSS_LOGISTIC_RECENT',
--- a/src/api/generatedTypes.ts
+++ b/src/api/generatedTypes.ts
@@ -626,6 +626,17 @@ export type CreateRejectedCorpusItemInput = {
   url: Scalars['Url'];
 };
 
+/**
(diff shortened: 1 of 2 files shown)


src/api/generatedTypes.ts
```diff
@@ -2104,6 +2137,7 @@ export enum Topics {
   Food = 'FOOD',
   Gaming = 'GAMING',
   HealthFitness = 'HEALTH_FITNESS',
+  Home = 'HOME',
```
jpetto, line 2140: i unfortunately think we need to manually remove this line. will give context in slack.

jpetto (changes requested):

src/api/generatedTypes.ts
jpetto: actually - i'll let @nina-py chime in here to confirm if we have to remove this line or not.

nina-py: It's ok to keep it in in the generated types as long as we're not using it in frontend code itself - this PR is waiting and has that change: https://github.com/Pocket/curation-admin-tools/pull/1212/files

jpetto (approved):
nina says we're good, so approved!

PR #1216, opened 2024-09-06
feat(corpus): [MC-1446] Add new 'Publisher Request' rejection reason

nina-py:
## Goal

Add the new rejection reason (`Publisher Request`) to Curation Tools frontend so that curators can start using it where necessary. 

Testing: I have verified it works after deploying the backend change to Dev.

## Follow-up

There are two more reasons available on the backend that we're NOT using on the frontend. These are `COMMERCIAL` and `PUBLISHER_QUALITY`. Should we add them, too?

## Screenshots

new-rejection-reason
 
## Reference

https://mozilla-hub.atlassian.net/browse/MC-1446

### DIFF
--- a/src/api/generatedTypes.ts
+++ b/src/api/generatedTypes.ts
@@ -1846,6 +1846,7 @@ export enum RejectionReason {
   Paywall = 'PAYWALL',
   PoliticalOpinion = 'POLITICAL_OPINION',
   PublisherQuality = 'PUBLISHER_QUALITY',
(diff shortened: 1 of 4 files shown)


jpetto (approved):
💡

jpetto:
> There are two more reasons available on the backend that we're NOT using on the frontend. These are `COMMERCIAL` and `PUBLISHER_QUALITY`. Should we add them, too?

good catch! ask this in slack?

PR #1218, opened 2024-09-26
MC-1341: Add "Filter by Topics" to prospecting page

katerinachinnappan:
## Goal

Add a "Filter by Topics" to Prospecting page. 

## Todos

- [x] deploy to dev

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-1341

## Implementation Decisions

* Only filter the first N loaded prospects on the page.
* Renamed `ScheduleDayFilter` to `DropDownFilter` & also refactored it a bit to make it more generic as it is now used by Scheduling page & Prospecting page.
* Took out the prospecting filtering logic from the Prospecting page & created a `getFilteredProspects` function where the filtering logic now lives to make it more cleaner & easier to test (no testing was done on this logic previously). 
* Renamed `ProspectPublisherFilter` to `ProspectFilters` as this component now houses publisher & topic filters & potentially can house more filters in the future.
* Slightly refactored `ScheduleDayFilterRow` to remove some duplicate code
* Renamed `ScheduleSummary` to `StoriesSummary` to be more generic as it is also used in Prospecting page now

## Demo

https://github.com/user-attachments/assets/93d20cfb-d4a9-4b48-bc8c-004d07f69edd


src/curated-corpus/components/ProspectFilters/ProspectFilters.tsx
```diff
@@ -5,9 +5,30 @@ import {
   Grid,
   Switch,
   TextField,
+  Typography,
 } from '@mui/material';
+import { Prospect } from '../../../api/generatedTypes';
+import { DropDownFilter } from '../DropDownFilter/DropDownFilter';
+// import { ProspectDropDownFilter } from '../ProspectDropDownFilter/ProspectDropDownFilter';
```
nina-py: Forgotten import?

src/curated-corpus/components/DropDownFilter/DropDownFilter.tsx
```diff
@@ -25,7 +26,15 @@ interface ScheduleDayFilterProps {
...
-  setFilters: React.Dispatch<React.SetStateAction<ScheduleDayFilterOptions>>;
+  setScheduleFilters?: React.Dispatch<
+    React.SetStateAction<ScheduleDayFilterOptions>
+  >;
+  /**
+   * Callback to set filters on the Prospecting Page
+   */
+  setProspectFilters?: React.Dispatch<
+    React.SetStateAction<ProspectFilerOptions>
+  >;
```
nina-py, line 37: I wonder if this could be simplified so we don't pass two similar, but not quite 100% compatible setter methods to the newly abstracted dropdown filter component. Ideally it should have one setter. A reasonably quick way to refactor this without having to rework much code would be to have a single `setFilters` function in the component props, but let it accept either `ProspectFilterOptions` or `ScheduleDayFilterOptions`.

src/curated-corpus/helpers/prospects.test.ts
```diff
@@ -226,4 +230,224 @@ describe('helper functions related to prospects', () => {
       },
     );
   });
+
+  describe('getFilteredProspects', () => {
+    const currentDate = DateTime.local();
```
nina-py, line 235: Love the increased test coverage! ⭐

src/curated-corpus/components/ProspectFilters/ProspectFilters.tsx
katerinachinnappan: @nina-py this seems outdated, I removed it in a recent commit!

nina-py (dismissed):
🔥

nina-py (approved):
Looks great! 🚀

PR #1220, opened 2024-10-03
fix(scheduling): [MC-1554] Fix same day scheduling from Schedule view

nina-py:
## Goal

Fix a bug described in the title of this PR.

- Updated validation rules for `scheduledDate` to evaluate today's timestamp that is the equivalent of the start of the day, and not any random time that occasionally prevented curators from schedulind.

- Updated validation messages to provide feedback to users when form validation fails.

- Removed an outdated prospect source from tests.

## Reference

https://mozilla-hub.atlassian.net/browse/MC-1554


katerinachinnappan (approved):
Yay! ✨🔥

PR #1222, opened 2024-10-30
MC-1556: Ensure ‘Coronavirus’ is not available as a Topic when adding an Item

katerinachinnappan:
## Goal

Upon adding, editing an item, ensure "Coronavirus" is not in the Topics list. For existing items with "Coronavirus" topic, the topic will be displayed as "N/A"

## Todos

- [x] deploy to dev

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-1556

### DIFF
--- a/src/curated-corpus/helpers/topics.test.ts
+++ b/src/curated-corpus/helpers/topics.test.ts
@@ -43,7 +43,7 @@ describe('helper functions related to topics', () => {
       expect(topics).to.be.an('array');
       // However, we're getting the entire list of topics here,
       // not just the ones stories belong to
-      expect(topics).to.have.lengthOf(17);
+      expect(topics).to.have.lengthOf(16);
 
       // And we expect to see the story topics, with the correct counts
       expect(topics).to.contain.deep.members([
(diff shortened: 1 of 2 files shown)


src/curated-corpus/helpers/topics.test.ts
```diff
@@ -94,7 +94,7 @@ describe('helper functions related to topics', () => {
 
     it('returns a full list of topics if requested', () => {
       const topics = getGroupedTopicData(data);
-      expect(topics).to.be.an('array').with.lengthOf(17);
+      expect(topics).to.be.an('array').with.lengthOf(16);
```
katerinachinnappan, line 97: should I get rid of the `'returns a list of topics without "Coronavirus" if requested'` test case as we don't show this topic?

nina-py: I think we should if this topic is no longer present anywhere in the rest of the codebase now!

nina-py (approved):

nina-py (approved):

PR #1223, opened 2024-11-02
MC-1583:"Fix Title” for English titles not capitalizing first word after single quote or colon

katerinachinnappan:
## Goal

Fix the capitalization rules for english titles:
* Capitalize the word "so" (remove as a stop word)
* Capitalize the first word (regardless if it is a stop word) after a quote: `'the day'` -> `'The Day'`
* Ensure there is a difference between a quote and apostrophe: `'the day' day'S` -> `'The Day' Day's`
* Capitalize word after a colon (:): '`the day' day's: day` -> `'The Day' Day's: Day`

Final result: `Amorim aims to 'prove something' at Manchester United after agreeing deal: read the day's 'news'` -> `Amorim Aims to ‘Prove Something’ at Manchester United After Agreeing Deal: Read the Day’s ‘News’`
 
## Todos

- [x] deploy to dev

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/MC-1583

### DIFF
--- a/src/_shared/utils/applyApTitleCase.ts
+++ b/src/_shared/utils/applyApTitleCase.ts
@@ -1,17 +1,24 @@
 export const STOP_WORDS =
   'a an and at but by for in nor of on or the to up yet';
 
+// Matches a colon (:) and 0+ white spaces following after
+// Matches 1+ white spaces
+// Matches special chars (i.e. hyphens, quotes, etc)
 export const SEPARATORS = /(:\s*|\s+|[-‑–—,:;!?()“”'‘"])/; // Include curly quotes as separators
 
 export const stop = STOP_WORDS.split(' ');
 
 /**
  * Format a string: Capture the letter after an apostrophe at the end of a
- * sentence (without requiring a space) or with a white space following the letter
+ * sentence (without requiring a space) or with a white space following the letter.
+ * Lowercase the captured letter & return the formatted string.
  * @param input
  * @returns {string}
  */
 export const lowercaseAfterApostrophe = (input: string): string => {
+  // matches a char (num or letter) right after an apostrophe,
+  // only if the apostrophe is preceded by a char & is followed
+  // by a space or end of the str.
   const regex = /(?<=\w)'(\w)(?=\s|$)/g;
 
   return input.replace(regex, (match, p1) => {
--- a/src/_shared/utils/applyApTitleCase.ts
+++ b/src/_shared/utils/applyApTitleCase.ts
@@ -66,8 +66,8 @@ export const applyApTitleCase = (value: string): string => {
         index > 0 &&
         (allWords[index - 1] === "'" ||
           allWords[index - 1] === '"' ||
-          allWords[index - 1] === '\u2018' ||
-          allWords[index - 1] === '\u201C');
+          allWords[index - 1] === '\u2018' || // Opening single quote ’
+          allWords[index - 1] === '\u201C'); // Opening double quote “
 
       if (
         index === 0 || // first word
--- a/src/_shared/utils/applyApTitleCase.ts
+++ b/src/_shared/utils/applyApTitleCase.ts
@@ -1,9 +1,23 @@
 export const STOP_WORDS =
-  'a an and at but by for in nor of on or so the to up yet';
+  'a an and at but by for in nor of on or the to up yet';
 
-export const SEPARATORS = /(\s+|[-‑–—,:;!?()“”"])/;
+export const SEPARATORS = /(:\s*|\s+|[-‑–—,:;!?()“”'‘"])/; // Include curly quotes as separators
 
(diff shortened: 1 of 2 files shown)


src/_shared/utils/applyApTitleCase.ts
```diff
@@ -1,10 +1,24 @@
 export const STOP_WORDS =
-  'a an and at but by for in nor of on or so the to up yet';
+  'a an and at but by for in nor of on or the to up yet';
 
-export const SEPARATORS = /(\s+|[-‑–—,:;!?()“”"])/;
+export const SEPARATORS = /(:\s*|\s+|[-‑–—,:;!?()“”'‘"])/; // Include curly quotes as separators
```
jpetto, line 4: for readability for future us, would it be feasible to add a comment here translating the different pieces of this regex?

i'm specifically curious about this part: `:\s*|\s+` i am not a regex wizard, but why is the `\s+` necessary? doesn't `\s*` cover that case?

src/_shared/utils/applyApTitleCase.ts
```diff
@@ -1,10 +1,24 @@
...
 export const stop = STOP_WORDS.split(' ');
 
+/**
+ * Format a string: Capture the letter after an apostrophe at the end of a
+ * sentence (without requiring a space) or with a white space following the letter
+ * @param input
+ * @returns {string}
+ */
+export const lowercaseAfterApostrophe = (input: string): string => {
+  const regex = /(?<=\w)'(\w)(?=\s|$)/g;
```
jpetto, line 15: a translation comment here would be great, too 😅

src/_shared/utils/applyApTitleCase.ts
```diff
@@ -1,10 +1,24 @@
...
 
 export const stop = STOP_WORDS.split(' ');
 
+/**
+ * Format a string: Capture the letter after an apostrophe at the end of a
+ * sentence (without requiring a space) or with a white space following the letter
+ * @param input
+ * @returns {string}
+ */
+export const lowercaseAfterApostrophe = (input: string): string => {
```
jpetto, line 14: could we add unit tests for this function?

src/_shared/utils/applyApTitleCase.ts
```diff
@@ -30,19 +44,36 @@ export const applyApTitleCase = (value: string): string => {
...
+  const result = allWords
     .map((word, index, all) => {
+      const isAfterColon = index > 0 && all[index - 1].trim() === ':';
+
+      const isAfterQuote =
+        index > 0 &&
+        (allWords[index - 1] === "'" ||
+          allWords[index - 1] === '"' ||
+          allWords[index - 1] === '\u2018' ||
+          allWords[index - 1] === '\u201C');
```
jpetto, line 63: would be good to add a comment to these unicode values as to what they represent

jpetto (changes requested):
functionally this looks great - hoping we can add some comments and targeted unit tests 😄

src/_shared/utils/applyApTitleCase.ts
katerinachinnappan: `:\s*` matches the `:` char & 0+ white spaces following the colon.
`\s+` matches only 1+ white spaces. 
So if we only had `:\s*` in the regex, it would only capture the `:` and any (or 0) white spaces trailing after. But we also want to capture just white spaces, and `\s+` allows you to do that.

jpetto: oh, i see - this is why i'm not a regex wizard. the `:\s*` and `\s+` are fully separate. i think i was reading it as `(:(\s*|\s+)|[-‑–—,:;!?()“”'‘"])/`

src/_shared/utils/applyApTitleCase.ts
```diff
@@ -30,19 +44,36 @@ export const applyApTitleCase = (value: string): string => {
...
-  // split by separators, check if word is first or last
-  // or not blacklisted, then capitalize
-  return value
-    .split(SEPARATORS)
+
+  // Split and filter empty strings
+  // Boolean here acts as a callback, evaluates each word:
+  // If it's a non-empty string, keep the word in the array;
+  // If it's an empty string (or falsy), remove from array.
+  const allWords = value.split(SEPARATORS).filter(Boolean); // Split and filter empty strings
```
jpetto, line 52: just out of curiosity, what is the `filter(Boolean)` guarding against?

src/_shared/utils/applyApTitleCase.ts
katerinachinnappan: i am no regex wizard either, https://regex101.com/ has my back whenever I need to test a regex 😅

jpetto (approved):
great work, and love the comments and tests!

PR #1244, opened 2025-06-26
HNT-689-migrate-to-webpack-5

jwalterss:
Fix webpack 5 compatibility issues and add local development
authentication bypass to enable proper local testing without OAuth
configuration dependencies.

Todos

Add webpack 5 browser polyfills for Node.js modules (crypto,
buffer, stream, util, assert)
Install required polyfill packages for webpack 5 compatibility
Fix runtime errors: PKCE sha256 function and process undefined
issues
Add OAuth callback route to handle authentication redirects
Implement local development authentication bypass with
REACT_APP_LOCAL_DEV flag
Add separate build script for local development (Node.js 17+
OpenSSL compatibility)
Ensure CI/CD builds work with Node.js 16 without legacy provider
Modifying Environment Variables

Added optional REACT_APP_LOCAL_DEV environment variable for local
development
When set to true, bypasses OAuth and provides mock
authentication
Not required for production/dev deployments - only for local
development convenience
Deployment

CI/CD builds use npm run build (works with Node.js 16 in Docker)
Local development can use npm run build:local (includes OpenSSL
legacy provider for Node.js 17+)
No deployment-specific changes required - existing OAuth flow
remains unchanged
Reference

Documentation:

https://webpack.js.org/migrate/5/
https://webpack.js.org/configuration/resolve/#resolvefallback
Implementation Decisions

Webpack 5 Polyfills: Added browser polyfills for Node.js modules
that were automatically included in Webpack 4 but removed in Webpack
5
Local Dev Bypass: Implemented optional authentication bypass to
allow local testing without OAuth provider configuration issues
Build Script Strategy: Created separate local build script to
handle Node.js 17+ OpenSSL compatibility while keeping CI/CD builds
clean
Backward Compatibility: All existing functionality preserved -
local dev bypass is entirely optional via environment variable


jpetto:
hey @jwalterss, finally getting some space to look at this...

i pulled this locally and started it up, and if i set `REACT_APP_LOCAL_DEV=true` in my `.env` file, i get the following error on seemingly any page that attempts to retrieve data:

> Context creation failed: Could not decode JWT

if i remove `REACT_APP_LOCAL_DEV=true` from my `.env` file, things function much as they did before - i get prompted for an SSO login and if that's successful data is returned.

am i doing something wrong? is this expected?

PR #1245, opened 2025-07-09
HNT-684-remove-manual-add-reason

jwalterss:
Goal

  Remove the requirement for curators to provide a reason when
  manually adding items to US & Germany surface schedules. 

  Todos

  - Remove "Reason for Manually Adding Item" UI section from
  ScheduleItemForm
  - Update validation to bypass manual schedule reason requirements
  - Clean up unused imports and variables
  - Test manual scheduling works without requiring reasons

  Tickets:

  - https://mozilla-hub.atlassian.net/browse/HNT-684 - Remove manual
  add reason requirement

  Implementation Decisions

  - UI-only approach: Made this a frontend-only change to avoid
  backend modifications. The GraphQL mutation still accepts reason
  parameters but they're no longer required by the UI.
  - Validation bypass: Modified the Yup validation schema to always
  return true for manual schedule reasons rather than removing the
  validation entirely, maintaining backward compatibility.
  - Preserved form structure: Kept the form's initial values and
  submission logic intact to ensure the backend continues to receive
  expected data structure (with empty/null reason values).


.env.example
```diff
@@ -0,0 +1,12 @@
+# Local development settings
+REACT_APP_LOCAL_DEV=false
+
```
katerinachinnappan, line 3: can you add this to the README of the repo? in the instructions for local dev, you can point to this `.env.example` file

jpetto: kinda cross posting here - i'd still like to know the purpose of `REACT_APP_LOCAL_DEV` (as [asked in 1244](https://github.com/Pocket/curation-admin-tools/pull/1244#issuecomment-3074561854)). it appears to not load data when set to `true`...

jwalterss: Apologies for letting this slip through the cracks. @katerinachinnappan I've updated the readme. @jpetto I reproduced that initial error and resolved it. When set to true, with a valid JWT token, the flow should work as normal.

jpetto (changes requested):
there are a lot of changes in this PR that also appear in #1244. let's simplify this PR to only target the changes outlined in the ticket - to remove the manual add reason - and leave #1244 to handle the larger chunk of work to upgrade to webpack 5.

jwalterss:
@jpetto What is the cleanest way to do that?

PR #1255, opened 2025-10-03
HNT-1105: hide ML removal reason & show set_diversity reason for removing section items

katerinachinnappan:
## Goal

Hide "ML" reason when removing section items as this reason is only used by ML processes. Show "SET_DIVERSITY" removal reason as it was hidden due to css.

## Todos

- [ ] deploy to dev

## Reference

Tickets:

- https://mozilla-hub.atlassian.net/browse/HNT-1105

### DIFF
--- a/src/curated-corpus/components/RemoveSectionItemForm/RemoveSectionItemForm.tsx
+++ b/src/curated-corpus/components/RemoveSectionItemForm/RemoveSectionItemForm.tsx
@@ -57,12 +57,16 @@ export const RemoveSectionItemForm: React.FC<
       onSubmit(values, formikHelpers);
     },
   });
+  // remove "ML" from list of reasons, its only used by ML process
+  const filteredReasons = Object.values(SectionItemRemovalReason).filter(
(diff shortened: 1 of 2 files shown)


src/curated-corpus/components/RemoveSectionItemForm/RemoveSectionItemForm.tsx
```diff
@@ -57,12 +57,16 @@ export const RemoveSectionItemForm: React.FC<
       onSubmit(values, formikHelpers);
     },
   });
+  // remove "ML" from list of reasons, its only used by ML process
+  const filteredReasons = Object.values(SectionItemRemovalReason).filter(
+    (reason) => reason !== SectionItemRemovalReason.Ml,
+  );
```
mmiermans, line 63: I wonder whether a cleaner solution would be to introduce two different sets of removal reasons, for example `EditorialSectionItemRemovalReason` that's extended with `ML` to form `SectionItemRemovalReason`. That way the special treatment of the `ML` can be re-used.

PR #1276, opened 2026-06-10
fix(curated-corpus): add Polish to the language picker

mmiermans:
## Problem
#1274 added Poland (`NEW_TAB_PL_PL`) as a scheduled surface, but Polish is missing from the language picker, so curators cannot set or see `PL` on corpus items and sections.

## Change
- Add Polish to the `languages` dropdown options (`definitions.ts`).
- Add `Pl = 'PL'` to the generated `CorpusLanguage` enum (`generatedTypes.ts`).

The enum value comes from the backing schema change in Pocket/content-monorepo#391, which is now deployed. `generatedTypes.ts` was updated by running `api:generate-types`.

### DIFF
--- a/src/api/generatedTypes.ts
+++ b/src/api/generatedTypes.ts
@@ -1348,6 +1348,8 @@ export type Mutation = {
    * Called when approving or rejecting a prospect into the corpus.
    */
   updateProspectAsCurated?: Maybe<Prospect>;
+  /** Updates a SectionItem's mutable fields (e.g. rank). */
+  updateSectionItem: SectionItem;
   /** Uploads an image to S3 for an Approved Item */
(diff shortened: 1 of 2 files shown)


src/api/generatedTypes.ts
```diff
@@ -2578,6 +2586,14 @@ export type UpdateLabelInput = {
...
 };
 
+/** Input data for updating a SectionItem's mutable fields */
+export type UpdateSectionItemInput = {
+  /** ID of the SectionItem. A string in UUID format. */
+  externalId: Scalars['ID'];
+  /** The updated rank of the SectionItem in relation to its siblings. */
+  rank: Scalars['Int'];
+};
+
```
mmiermans, line 2596: Generated from dev schema.

jpetto (approved):

